PR #37, opened 2013-09-27
adding read_exactly method and default gw routing.

crotsos:
Adding a read_exactly method in the channel module definition. 

Modifying the routing method to select the default gw in case no routing entry is matched.


samoht:
I believe that `read_exactly` is now cherry-picked.

PR #98, opened 2015-01-13
Added 'error_message' functions

talex5:
Needed for https://github.com/mirage/mirage/pull/346


talex5:
I've also added some useful ICMP error reporting to replace the old `ICMP unknown ty 3` message.

PR #102, opened 2015-01-27
Support io-page 1.3.0+

talex5:


samoht:
@avsm can we have a release with this changeset?

PR #104, opened 2015-02-05
Arp state machine

yomimono:
Improve ARP handling for IPv4 as follows:

* Create ARP cache entries for unsolicited (gratuitous) ARP replies.  Previously, receipt of these packets was noted in the console output but no entries were made in the ARP cache.
* Time out entries in the ARP cache after 60 seconds.
* If no reply to an ARP request is received within 1.5 seconds, send another ARP request.  A total of 3 ARP requests will be sent before giving up.
* In Ipv4, throw exception No_route_to_destination_address if a local address could not be resolved.  Previously, this exception was thrown only if a connection attempt was made to a non-local unicast address and no gateways were configured; it will now also be thrown if a gateway is configured but does not respond to ARP requests, or a local unicast address does not respond to ARP requests.  (Alternate suggestions for handling this are welcome.)

### DIFF
--- a/lib/arpv4.ml
+++ b/lib/arpv4.ml
@@ -18,7 +18,7 @@
 open Lwt
 open Printf
 
-module Make (Ethif : V1_LWT.ETHIF) = struct
+module Make (Ethif : V1_LWT.ETHIF) (Time : V1_LWT.TIME) = struct
   type arp = {
     op: [ `Request |`Reply |`Unknown of int ];
     sha: Macaddr.t;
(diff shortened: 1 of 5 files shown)


yomimono:
This PR also needs https://github.com/mirage/mirage/pull/362 to generate usable unikernels.

lib/arpv4.ml
```diff
@@ -175,18 +197,30 @@ module Make (Ethif : V1_LWT.ETHIF) = struct
...
+        Hashtbl.remove t.cache ip; (* new attempts to resolve should not be referred
+                                        to the previous sleeping thread *)
+        match counter with 
+        | 0 -> return None
+        | n ->
+          let response, waker = MProf.Trace.named_wait "ARP response" in
+          Hashtbl.add t.cache ip response;
+          Hashtbl.add t.pending ip waker; 
+          output_probe t ip >>= fun () -> 
+          Time.sleep probe_repeat_delay >>= fun () ->
```
talex5: If I'm reading this right, we always wait for `probe_repeat_delay` after the ARP request is sent, even if we get a reply sooner. I think this means that the first user to request an address will always have to wait (other callers may get the answer from the cache even while the first caller is still waiting).

Also, any caller waiting on the initial cached `response` will keep waiting if it fails, even if there is a successful retry.

I'd suggest putting the retry logic on the inside of the loop, so the entry in `t.cache` represents the result of the whole lookup-with-retry operation, not just the result of the current try. Then you don't have to modify the cache on each retry.

I haven't tested it, though, so could be talking nonsense...

avsm:
Saving this for a post 2.2.3 release after more review/testing.

samoht:
Superseded by #118

PR #108, opened 2015-02-16
Expose universal IP adresse converters

samoht:
Should be merged with https://github.com/mirage/mirage/pull/361


samoht:
Actually, this could be merged now, and mirage/mirage#361 later.

PR #111, opened 2015-03-01
Validate TCP and UDP (unless 0) checksums

hannesm:
while reading through the code, I removed some dead code (Profiler/Sliding_window/parts of Tcp.Pcb).
I also implemented validation of TCP and UDP checksums (because I think it is wrong to just not care about them).
While doing that, I had to modify the signature of `checksum` (see https://github.com/mirage/mirage/pull/369) -- because when receiving packets, the UDP/TCP layer only gets the IP payload, not the complete ethernet frame.

This will of course be slightly slower than the old code (which did not try to validate checksums), but more correct.


hannesm:
I also couldn't stand to have magic numbers spread over the library, thus I implemented `protocol_to_int` and `int_to_protocol` in Wire_structs, which deals with these conversions (could very well be extended for matches such as https://github.com/mirage/mirage-tcpip/blob/master/lib/ipv6.ml#L465)

avsm:
This will need some careful validation for VM<->VM traffic, since Xen disables checksum generation in some cases there (advertised via a Xenstore feature flag).

hannesm:
@avsm interesting... is there some documentation of this feature?

avsm:
None apart from this old thread on the same topic: https://lists.cam.ac.uk/pipermail/cl-mirage/2013-December/msg00006.html

talex5:
There's some info about Xen checksums here: http://lists.xen.org/archives/html/xen-devel/2013-12/msg00884.html

hannesm:
ic. so checksum offloading should be configurable at `connect` time (do we know at that point whether we're running on xen? this most likely needs support in the mirage tool, doesn't it?) -- and involves (enables or disables) both Rx and Tx checksum computations!?

avsm:
It needs support from the Netif to check for the feature flags at device connection time; there is probably a connection to mirage/mirage#286 which adds checksum offload support.

samoht:
ping?

hannesm:
the first 3 commits have already been merged, the whole TCP/IP checksum validation (including potential offloading) needs to be done (depending on xen), https://github.com/mirage/mirage/pull/286 seems to be part of it. closing this PR since it is completely out of date.

PR #112, opened 2015-03-02
Minor cleanups

hannesm:
after some discussions today, I applied the cleanups in this branch while not doing anything with checksumming (apart from moving the pattern match around).

this is a completely semantics and API preserving change, which should be straightforward to merge!


samoht:
lgtm

avsm:
I think @balrajsingh manually hooks in the profiler for testing, and the sliding_window should be used (as opposed to getting cleaned out).  Could you separate the removal of those from the (very good) wire_format changes?

hannesm:
Window implements the invariants and code which is partially implemented in Sliding_window, afaics...

avsm:
The goal is to reduce the complexity of Window by pulling out data structures into smaller, standalone components such as `Sliding_window` (which doesn't have all the other bits of the `Window.t` type).  It also helps with functional testing to ensure that (e.g.) wrapping behaviour of `Sliding_window` is correct, which isn't easily done inside the massive `Window` module.

hannesm:
I completely agree to the goal, but would start with the current (working) code in `Window`, and migrate this into a purely functional interface.. esp since `Sliding_window` is hooked up in the build, I expected it to be used somewhere.. but it only managed to confuse me while reading the library.

in any case, force-pushed to omit the removal of both Profiler and Sliding_window

avsm:
Fair point -- I'm happy for the modules to be removed from the build

hannesm:
done in 38cf180...

PR #114, opened 2015-03-04
Filter incoming frames by MAC address in Ethif

MagnusS:
Only accept frames sent to broadcast or local MAC address. 

Partially fixes #113, as packets intended for other hosts are not sent to the IP layer.

PR #115, opened 2015-03-09
release 2.2.3

avsm:
* Add ICMPv6 error reporting functions (#101)
* Add universal IP address converters (#108)
* Add `error_message` functions for human-readable errors (#98)
* Improve debug logging for ICMP Destination Unreachable packets.
* Filter incoming frames by MAC address to stop sending unnecessary RSTs. (#114)
* Unhook unused modules `Sliding_window` and `Profiler` from the build. (#112)
* Add an explicit `connect` method to the signatures. (#100)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,10 +1,11 @@
-2.2.3 (2015-03-04):
+2.2.3 (2015-03-09):
 * Add ICMPv6 error reporting functions (#101)
 * Add universal IP address converters (#108)
 * Add `error_message` functions for human-readable errors (#98)
 * Improve debug logging for ICMP Destination Unreachable packets.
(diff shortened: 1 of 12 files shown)


hannesm:
hmm, this `connect` feels slightly awkward.. shouldn't we strive for modularity and thus being able to run an ip stack without tcp? in order to do so, `connect` of tcp should register the input handler in the ip stack... instead of (as now) having it hardcoded in ip (and always require tcp/udp stacks)!? (most likely a feature rarely used, maybe for 3.0..)

avsm:
@hannesm yes, I agree with that sentiment, and it'll be very useful for functional testing as well.  For a post 2.2.3 release though, as this is needed for the `connect` removal merge in the short term.

PR #117, opened 2015-03-11
check length of received ethernet frame (to avoid cstruct exceptions)

hannesm:
provide ethertype_to_protocol / protocol_to_ethertype conversion
cleanup code

should be straightforward. please let me know if you consider this code to be less readable

### DIFF
--- a/lib/ethif.ml
+++ b/lib/ethif.ml
@@ -46,16 +46,14 @@ module Make(Netif : V1_LWT.NETWORK) = struct
       (* minimum payload is 46 + source + destination + type *)
       match Macaddr.of_bytes (Wire_structs.copy_ethernet_dst frame) with
       | Some frame_mac when of_interest frame_mac ->
-        begin
-          let payload = Cstruct.shift frame Wire_structs.sizeof_ethernet
-          and ethertype = Wire_structs.get_ethernet_ethertype frame
-          in
-          match Wire_structs.ethertype_to_protocol ethertype with
-          | Some `ARP -> arpv4 frame
-          | Some `IPv4 -> ipv4 payload
-          | Some `IPv6 -> ipv6 payload
-          | None -> return_unit (* TODO default etype payload *)
-        end
+        let payload = Cstruct.shift frame Wire_structs.sizeof_ethernet
+        and ethertype = Wire_structs.get_ethernet_ethertype frame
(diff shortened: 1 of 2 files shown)


mor1:
looks good to me, assuming travis passes.

is there a reason not to use Cstruct's `cenum` to declare the constants and associated convertors in `wire_structs.ml`?  cf. eg., <https://github.com/mirage/ocaml-dns/blob/master/lib/packet.ml>

hannesm:
cenum produces variants -- thus they need to be qualified -- whereas I use polyvars here.. I'm still not decided what the best way [tm] to do this is... most likely will switch over to cenum and variants, for the simple sake of speed...

hannesm:
@mor1 see 2499b57 :)

mor1:
cool :)

hannesm:
hmm, actually I like 1e3e291 much better.. what do you think? separating the parser out (should actually go into a different module than Wire_structs, shouldn't it?)

mor1:
tbh i think there's some general stuff to do about cleaning up parsing in all our protocol stacks, cf discussions about CPS with @pqwy. certainly separating parsing from input processing makes sense to me (and would probably allow eg `pcap-format` to more easily reuse it too).

hannesm:
yes, this is a start. we can rewrite it later, once we know how the generic parsing interface should look like..

samoht:
I like it

hannesm:
ping? any opinion (this change should be straightforward, and not change semantics)

samoht:
Can you rebase your patch? I'm happy to merge once done.

hannesm:
rebased

samoht:
Hum, the ARP tests are not working very well with your changes ...

hannesm:
`parse_ethernet_frame` is now more picky, it requires each packet to be at least 60 bytes in size (as [specified here](https://en.wikipedia.org/wiki/Ethernet_frame)).  Our ARP implementation does not pad frames to that size, but sends out ethernet frames of size 42 (and relies on the NIC to pad).

It seems to be sensible to let the NIC pad the data, thus I relaxed the check in 26c9311.

hannesm:
@samoht now travis is happy and the tests are passing

samoht:
Great, thanks!

PR #118, opened 2015-03-17
Arp improvements

nojb:
I find the code in `arpv4.ml` hard to follow and reason about.  See for example the discussion in #104.  In particular, it is hard to modify the code without introducing race conditions.

This patch tries to improve this state of affairs.  Along the way I incorporated @yomimono's patches in #104.  The first two commits contain the basic changes:  introduce a datatype `entry` to stop representing illegal states (right now one needs to be careful to keep `t.pending` and `t.cache` in sync which can easily lead to race conditions).  The rest of the commits just incorporate the changes of #104.

In order to timeout stale entries a single background thread is spawned on creation (see `tick`) which cleans up every 60s.

The only change that could potentially affect current users is that ARP lookups now timeout after `1.5s * 3`.  Before, ARP lookups would just linger forever if no answer was received.  The timeout values were taken from #104, where it refers to RFC 5227.

The mirage tool requires a small patch to work since the proposed ARPV4 (and hence IPV4) is a functor of `CLOCK` and `TIME`, so the generated code in `main.ml` needs to be modified accordingly.

I tested the patch with some of the examples of `mirage-skeleton` without encountering any problems.  Thoughts, ideas, comments ?

/cc @yomimono

### DIFF
--- a/lib/arpv4.ml
+++ b/lib/arpv4.ml
@@ -30,13 +30,12 @@ module Make (Ethif : V1_LWT.ETHIF) = struct
   (* TODO implement the full ARP state machine (pending, failed, timer thread, etc) *)
 
   type entry =
-    | Pending
+    | Pending of Macaddr.t Lwt_condition.t
     | Confirmed of Macaddr.t
 
   type t = {
     ethif : Ethif.t;
     cache: (Ipaddr.V4.t, entry) Hashtbl.t;
-    pending: (Ipaddr.V4.t, Macaddr.t Lwt_condition.t) Hashtbl.t;
     mutable bound_ips: Ipaddr.V4.t list;
   }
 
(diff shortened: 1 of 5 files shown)


talex5:
Seems reasonable to me. A few notes:
- I find `Not_found` exceptions can be frustrating (easy to forget to catch, not clear where they come from)... maybe use a variant here? Though this isn't part of the public interface, so not a big problem. Or, an `ARP_timeout` exception would at least make it obvious what the cause was.
- This keeps ticking once per second even when the cache is empty. Could this prevent sleeping? Probably worth it to keep the code simple, though.
- I think @yomimono is working on some tests for the ARP code. Would be good to try it with them.

yomimono:
Unlike my patch, this does indeed appear to pass the tests I could think of for reasonable ARP behavior. :)

samoht:
I agree with @talex5 that [Not_found](https://github.com/mirage/mirage-tcpip/pull/118/files#diff-09e69c255360377e7089587d4c88a421R226) is not a terrible idea here.

dbuenzli:
> Not_found is not a terrible idea here

Not only here. It is *always* a terrible idea.

nojb:
By popular demand, I switched `Not_found` for `ARP_timeout`.  Thanks!

dbuenzli:
> By popular demand, I switched Not_found for ARP_timeout. Thanks!

A timeout doesn't feel like something exceptional. You should rather change the `query` signature from : 

```ocaml
val query: t -> Ipaddr.V4.t -> Macaddr.t Lwt.t
```

to 
```ocaml
val query: t -> Ipaddr.V4.t -> [ `Ok of Macaddr.t | `Timeout ] Lwt.t
```

nojb:
Agreed.  Thanks!

Issue #107, opened 2015-02-08
The stack should stop transmitting after receiving a RST


samoht:
This is a typical trace in synjitsu:

```
3134 1814.067582  192.168.2.1 -> 192.168.2.140 TCP 78 63986→80 [SYN] Seq=0 Win=65535 Len=0 MSS=1460 WS=32 TSval=2076397423 TSecr=0 SACK_PERM=1
3135 1815.074428  192.168.2.1 -> 192.168.2.140 TCP 78 [TCP Retransmission] 63986→80 [SYN] Seq=0 Win=65535 Len=0 MSS=1460 WS=32 TSval=2076398423 TSecr=0 SACK_PERM=1
3136 1815.076512 192.168.2.140 -> 192.168.2.1  TCP 62 80→63986 [SYN, ACK] Seq=0 Ack=1 Win=65535 Len=0 MSS=1460 WS=4
3137 1815.076575  192.168.2.1 -> 192.168.2.140 TCP 54 63986→80 [ACK] Seq=1 Ack=1 Win=262144 Len=0
3138 1815.076676  192.168.2.1 -> 192.168.2.140 HTTP 72 GET / HTTP/1.0
3139 1815.080134 192.168.2.140 -> 192.168.2.1  TCP 71 [TCP segment of a reassembled PDU]
3140 1815.080180  192.168.2.1 -> 192.168.2.140 TCP 54 63986→80 [ACK] Seq=19 Ack=18 Win=262112 Len=0
3141 1815.080304  192.168.2.1 -> 192.168.2.140 TCP 54 63986→80 [FIN, ACK] Seq=19 Ack=18 Win=262112 Len=0
3142 1815.081467 192.168.2.140 -> 192.168.2.1  HTTP 135 HTTP/1.1 200 OK
3143 1815.081503  192.168.2.1 -> 192.168.2.140 TCP 54 63986→80 [RST] Seq=19 Win=0 Len=0
3144 1815.081512 192.168.2.140 -> 192.168.2.1  TCP 60 [TCP Dup ACK 3142#1] 80→63986 [ACK] Seq=99 Ack=19 Win=262140 Len=0
3145 1815.081523  192.168.2.1 -> 192.168.2.140 TCP 54 63986→80 [RST] Seq=19 Win=0 Len=0
3146 1815.084058 192.168.2.140 -> 192.168.2.1  TCP 60 80→63986 [ACK] Seq=99 Ack=20 Win=262140 Len=0
[...]

PR #119, opened 2015-03-24
fix channel resource leak on close of underlying flow; add exceptions to channel

yomimono:
In investigating https://github.com/mirage/mirage/issues/371 , I discovered a resource leak in `channel.ml`; included in this patch set is a fix for this issue.  

I've added some exceptions in an attempt to express the possibility for failure (previously unacknowledged in the exposed interface). This is an attempt at compromise between the more unified error handling likely to be in use in the future, and my own desire to have something better now instead of something perfect later.

I've added some extremely minimal unit tests for channel to the (new) lib_test directory.  They require mirage-flow and ounit, both of which are set as depopts in `opam`.


opam
```diff
@@ -31,6 +31,8 @@ depends: [
   "mirage-net-unix" {>= "1.1.0"}
   "ipaddr" {>= "2.2.0"}
   "mirage-profile"
+  "mirage-flow" {test}
+  "ounit2" {test}
```
samoht: Should be `ounit`

samoht:
Great! One thing we were discussing with @avsm was to extract that module from `mirage-tcpip` completely (ie. have a `mirage-channel`). We can do this later though.

channel/channel.ml
```diff
@@ -50,66 +52,82 @@ module Make(Flow:V1_LWT.FLOW) = struct
...
+         exceptions *)
       t.ibuf <- Some buf;
       return_unit
-    | `Error _ | `Eof ->
-      fail Closed
+    | `Error e ->
+      fail (Read_error (Flow.error_message e))
+    | `Eof ->
+      (* close the flow before throwing exception; otherwise it will never be
+         GC'd *)
```
talex5: This sounds like it might be another bug. Why doesn't the flow get GC'd eventually? Can we close forgotten flows automatically using `Gc.finalise` or something (as a last resort, logging a warning)?

channel/channel.ml
```diff
@@ -25,6 +25,10 @@ module Make(Flow:V1_LWT.FLOW) = struct
   type +'a io = 'a Lwt.t
   type 'a io_stream = 'a Lwt_stream.t
 
+  exception End_of_file (* at least one user understands this exception *)
```
avsm: Not sure what this comment means -- do you mean one external user is already expecting this exception to be thrown?

channel/channel.ml
yomimono: The flow provided by `tcp/flow.ml` (and `tcp/pcb.ml`) doesn't do anything to clean itself up on EOF; rather it relies on the state machine logic (`tcp/state.ml`) to call `close` when logic dictates that the communications are over.  I suspect this particular case is related to https://github.com/mirage/mirage-tcpip/issues/107 (we don't close Established connections when we receive an RST).

channel/channel.ml
yomimono: Yep. https://github.com/mirage/mirage-http/blob/master/lib/HTTP_IO.ml#L40

avsm: Aha, good catch! I think that might be `Pervasives.End_of_file`, so it still wouldn't be caught by defining another `End_of_file` here in Channel (exceptions are not like polymorphic variants -- if you define them twice in two different modules, they are two different exceptions for the purposes of catching them).  I suspect that `End_of_file` leaked from the socket backend (still needs to be confirmed).  This basically reinforces the "exceptions need to disappear from these interfaces" argument, but this fixup en route is just fine.

channel/channel.ml
talex5: I see `Flow` currently has `type t = Pcb.t`. If we instead wrapped it as `type t = Flow of Pcb.t` then we could attach a finaliser to it. If that got called, it would mean that neither `write` nor `close` could ever be called again and we could close the flow automatically (with a warning), I think.

channel/channel.ml
talex5: That code is doing two horrible things: using an exception to represent the (non-exceptional, should-be-handled) case of end of file, and then converting it to the magic value `""` to mean the same thing.

The cohttp docs say:

> It returns an empty string if EOF *or some other error condition* occurs on the input channel

Would be good to standardise on something a bit saner everywhere...

channel/channel.ml
```diff
@@ -195,12 +213,15 @@ module Make(Flow:V1_LWT.FLOW) = struct
...
+    Flow.writev t.flow l >>= function
+    | `Ok () -> Lwt.return_unit
+    | `Error (e : Flow.error) -> fail (Write_error (Flow.error_message e))
+    | `Eof -> fail (End_of_file)
 
   let close t =
-    flush t
-    >>= fun () ->
-    Flow.close t.flow
+    try_lwt
```
nojb: You could use
```ocaml
Lwt.finalize (fun () -> flush t) (fun () -> Flow.close t.flow)
```
to avoid adding the `lwt.syntax` dependency (since `camlp4` is on its way out).

samoht:
Supersed by #127

channel/channel.ml
samoht: I'm a bit afraid that adding finalisers on every flow will put a lot of pressure on the GC.

channel/channel.ml
samoht: agreed ... I think keeping the same (crazy) semantics in that PR is useful though,

PR #121, opened 2015-03-30
adjust assert to allow trailing bytes

hannesm:
see issue https://github.com/mirage/mirage-net-xen/issues/24

### DIFF
--- a/lib/ipv4.ml
+++ b/lib/ipv4.ml
@@ -184,7 +184,6 @@ module Make(Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = stru
     let ihl = (Wire_structs.Ipv4_wire.get_ipv4_hlen_version buf land 0xf) * 4 in
     let src = Ipaddr.V4.of_int32 (Wire_structs.Ipv4_wire.get_ipv4_src buf) in
     let dst = Ipaddr.V4.of_int32 (Wire_structs.Ipv4_wire.get_ipv4_dst buf) in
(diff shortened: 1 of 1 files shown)


avsm:
@samoht it's now very hard to release tcpip since you seem to have the lock on it and have had it for a while... what needs to happen to merge this and unbreak ARM?

avsm:
This one needs to merge, but I think it shouldn't be an assert.  We should raise a proper exception or use some other functional callback to indicate we dropped the packet (in fact, the latter would be good: a `?drop_action` that would be a logger or just an identity function to discard it)

hannesm:
I think due to it being very partial (there are lots of other potential out-of-bounds accesses, such as the split a few lines above), the assert can just be dropped. It would be good to have an interface which does not throw exceptions in a future tcp/ip release.

hannesm:
@samoht @avsm if you want to release TCP/IP + mirage, I'd be happy to back out https://github.com/mirage/mirage/pull/390 (and wait to get conduit/tls/entropy for 2.4.7 then).

avsm:
Let's not roll anything back. I'll cut a point release of tcpip with this fixed as a branch.

avsm:
v2.3.1 released.  Everyone ok with a silent drop of the malformed frame in #122 instead of an assert?

samoht:
We should at least print a debug message on the console when this happens.

avsm:
No, we shouldn't.  All such messages need to go to some logging service and not to the console (which is a denial of service very easily).

hannesm:
this is superseeded by changes in 2.3.1 / 2.4.1

PR #123, opened 2015-04-21
Merge 2.4.0 and 2.3.1

samoht:


avsm:
Could you also merge this with #124

koleini:
I still suggest truncating IP packets into their actual size:

https://github.com/mirage/mirage-net-xen/issues/24#issuecomment-90519636

hannesm:
I actually suggest that the only place to truncate / remove trailing bytes is the ethernet layer... all layers above should be picky about checking trailing bytes (and underflows)!

avsm:
Agreed.

hannesm:
only problem is that an ethernet frame does not contain its payload length :/ thus, I'd really do the cutting of trailing bytes in mirage-net-xen..

samoht:
@MagnusS do we have a way to test that this bug will never pop-up (using you vnetif implementation and injecting random frames)? Would be very good if you can add a skeleton to test this kind of things more widely.

MagnusS:
Yes, that would be possible. One way to do it would be to write a module that wraps the Vnetif [backend](https://github.com/MagnusS/mirage-vnetif/blob/master/lib/basic_backend.mli) and randomly adds trailing bytes in calls to write/writev. We could then run something like iperf between two unikernels to verify that tcpip still works (as in this [example](https://github.com/MagnusS/mirage-vnetif/blob/master/examples/iperf_self/iperf_self.ml)).

PR #124, opened 2015-04-21
bump working version to 2.4.1, include the `opam` file

avsm:
OPAM being out of sync was reported by Ania Skorupa


samoht:
Do we really want to duplicate the version field in `oasis` and `opam`?

dbuenzli:
I'm not sure I understand the advantage of having the version number in the repo's `opam` file *at all*.

avsm:
To deal with constraints. I can then put an upper bound on incompatible packages and use this pin to get over them.

But crucially, it should also never have the *wrong* version number, as it did before this patch.  That wreaks havoc.

samoht:
I removed the version number in the opam file, people can pin to a specific version by doing:

```
opam pin add tcpip.<version> --dev
``

PR #126, opened 2015-04-23
Synjistu

samoht:
Cleaner implementation of #125

- Remove the ARP hack
- Use a global KV store (whose implementation can be passed as functor argument)
- Do not allow to listen on (-1) to listen on all ports (see next point)
- Synjitsu always write SYN it doesn't know in xenstore, even if there is no listeners for the given port


tcp/KV.mli
```diff
@@ -0,0 +1,34 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+module type S = sig
```
avsm: A little bit of documentation about what this module is for would be useful...

tcp/options.ml
```diff
@@ -16,6 +16,8 @@
 
 (* TCP options parsing *)
 
+open Sexplib.Std
```
avsm: Sexplib.Conv is better than Sexplib.Std isnt it?

samoht: better in what sense? I just want to bring the `sexp_of_int` and other basic converters in scope.

samoht:
There are few tricky (but self-contained) patches who might need some review before merging:

- why do we call `Flow.listen` inside `Flow.connect`?
- why do we reply to all packets, even if the dst IP is not us?
- For synjitsu to work, we need a way to set the stack in promiscious mode (instead of being always on, as in 44ec1e2)
- I think we need to update the ARP table properly when reading from xenstore, otherwise I got some `No_route_to_host` exceptions

samoht:
@yomimono true. But would be great to keep a branch with these changes if anyone want to replicate the papers' settings.

yomimono:
@samoht what do you think of closing the PR but pushing the branch to `mirage-tcpip`?

samoht:
Done.

PR #129, opened 2015-04-30
Add vnetif tests for connect and iperf + backend for trailing bytes

MagnusS:
Adds two new tests:

- test_connect connects two network stacks and sends a short hello world string
- test_iperf runs iperf between two network stacks

Both tests are run with the basic mirage-vnetif backend, as well as a new backend that adds random trailing bytes.

Depends on [mirage-vnetif](https://github.com/MagnusS/mirage-vnetif)


avsm:
superb!

MagnusS:
Mirage-vnetif is submitted to opam, but not merged yet ([PR](https://github.com/ocaml/opam-repository/pull/4034)) - so Travis will fail for now

samoht:
Can you `ocp-indent` the files?

yomimono:
I can only run the tests if tcpip is already installed and has made `dlltcpip_stubs` available somewhere -- otherwise, `test.byte` fails to load a shared library:

```
$ make test
ocaml setup.ml -build 
W: Cannot find source file matching module 'tcp' in library tcp
Finished, 0 targets (0 cached) in 00:00:00.
Finished, 248 targets (226 cached) in 00:00:02.
ocaml setup.ml -test 
Fatal error: cannot load shared library dlltcpip_stubs
Reason: dlltcpip_stubs.so: cannot open shared object file: No such file or directory
W: Test 'test' fails: Command '/home/user/mirage-tcpip/_build/lib_test/test.byte' terminated with error code 2
E: Failure("Tests had a 100.00% failure rate")
make: *** [test] Error 1
```

Presumably we'd want to run against the version of this library that's just been built, rather than whatever's in ~/.opam?

yomimono:
(also I'm very excited about the presence of these tests)

samoht:
> `Fatal error: cannot load shared library dlltcpip_stubs`

This is because the tests needs to be compiled with `-custom`, see https://github.com/mirage/ocaml-cstruct/blob/master/_oasis#L75

lib_test/vnetif_common.ml
```diff
@@ -0,0 +1,54 @@
...
+let create_stack c backend ip netmask gw =
+    or_error "backend" S.V.connect backend >>= fun netif ->
+    (* Printf.printf (Printf.sprintf "Connected to backend with mac %s" (Macaddr.to_string (S.V.mac netif))) *)
+    or_error "ethif" S.E.connect netif >>= fun ethif ->
+    or_error "ipv4" S.I.connect ethif >>= fun ipv4 ->
+    or_error "udpv4" S.U.connect ipv4 >>= fun udpv4 ->
+    or_error "tcpv4" S.T.connect ipv4 >>= fun tcpv4 ->
+    let config = {
+        V1_LWT.name = "stack";
+        V1_LWT.console = c;
```
avsm: Just one `V1_LWT` needed here; rest are redundant in the record as it is inferred.

MagnusS:
Thanks! I've changed the following:
- Added -custom to _oasis and updated
- Cleaned up the code in common.ml based on Thomas' suggestions 
- Removed Lwt.syntax dependency
- Removed redundant V1_LWT
- ocp-indent'ed everything

samoht:
Yay!

PR #130, opened 2015-04-30
ARP, IP and TCP improvements

samoht:
* Allow the ARP stack to be configured in promiscuous mode
* Process only IP packets with the right destination IP
* Make TCP options, sequences, and wires serializable
* Do not call `Random.self_iniit` on every new connect


CHANGES
```diff
@@ -1,3 +1,10 @@
+2.4.3 (trunk)
+* Allow the ARP stack to be configured in promiscuous mode
+* Process only IP packets with the right destination IP
+* Make TCP options, sequences, and wires serializable
+* Do not call `Random.self_iniit` on every new connection
```
avsm: s/iniit/init

lib/ethif.ml
```diff
@@ -72,7 +75,10 @@ module Make(Netif : V1_LWT.NETWORK) = struct
 
   let connect netif =
     MProf.Trace.label "ethif.connect";
-    return (`Ok { netif })
+    return (`Ok { netif; promiscuous = false })
 
   let disconnect _ = return_unit
+
+  let enable_promiscuous_mode t = t.promiscuous <- true
+  let disable_promiscuous_mode t = t.promiscuous <- false
```
avsm: there also needs to be a query function for the current promiscuous state

tcp/pcb.ml
```diff
@@ -18,6 +18,8 @@
 open Lwt
 open Printf
 
+let () = Random.self_init ()
```
infidel: Just wondering, could it matter if this Random were a different module from the Random that is passed as a parameter to the Make functor? I guess it must have been parameterised for a reason.

samoht: good point ...

samoht:
Thanks all, I've tried to take the comments into account. The main change is the IP filtering, hopefully it is done at the right place now but I'm not sure it is doing the right thing ... need some tests :-)

lib/arpv4.ml
```diff
@@ -103,6 +103,8 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
     | Not_found ->
       Hashtbl.replace t.cache ip (Confirmed (expire, mac))
 
+  let add t ip mac = notify t ip mac
```
hannesm: should a manually added arp entry really time out? in UNIX, such static arp entries do not timeout...

samoht: the idea is to call that function on every packet. but you are right, I should probably make the name a bit more explicit and speak about timeout in the docs. And maybe expose a `add_static` function.

lib/arpv4.mli
```diff
@@ -28,6 +28,9 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) : sig
       [t]. *)
   val create: Ethif.t -> t
 
+  val add: t -> Ipaddr.V4.t -> Macaddr.t -> unit
+  (** Add an entry in the ARP table. *)
```
hannesm: what is the failure behaviour? I'd add a policy what to do if there's already an entry for the given ipaddr, `(?if_exists : Replace | Error | Drop)`

lib/arpv4.ml
hannesm: also, static arp entries shouldn't be overwritten by incoming packets (to prevent arp spoofing)

samoht: that's a good point ...

samoht: SYNJITSU needs to do this for packets read from xenstore. I guess that's not a good idea to do it in the general case and some care should be taken even when it's read from xenstore to avoid spoofing.

samoht:
Closing for now on, I'll re-open with an improved version later.

PR #134, opened 2015-05-06
Expose Arpv4.Make functor; require arp argument for Ipv4.connect

yomimono:
Pull ARP out of ipv4 and make it externally usable.  Require explicit arp.t argument for Ipv4.connect .

Needs to be merged in tandem with https://github.com/mirage/mirage/pull/400 .


_oasis
```diff
@@ -133,6 +133,17 @@ Library "ethif-unix"
                       lwt,
                       lwt.unix
 
+Library "arpv4-unix"
```
samoht: Could we easily expose a `arpv4` library without unix dependencies?

lib/wire_structs.ml
```diff
@@ -45,6 +45,30 @@ module Ipv4_wire = struct
...
+    uint8_t tha[6];
+    uint32_t tpa
+  } as big_endian
+
+  cenum op {
+    Op_request = 1;
+    Op_reply
+    } as uint16_t
+
+end
```
samoht: I'm in favour of breaking the consistency of the library and put ARP wire format directly in `arp.ml` (or in `arp_wire.ml`).

_oasis
yomimono: Good catch; thanks! Fixed.

_oasis
```diff
@@ -40,6 +40,14 @@ Library ethif
...
   BuildDepends:       tcpip,io-page,mirage-types,ipaddr,cstruct,lwt
 
+Library arpv4
+  CompiledObject:     best
+  Path:               lib
+  Findlibparent:      tcpip
+  Findlibname:        arpv4
+  Modules:            Arpv4
+  BuildDepends:       tcpip,io-page,mirage-types,ipaddr,cstruct,lwt
+
```
samoht: I think you also need to remove `Arp` from the list of packages in `ipv4`

_oasis
```diff
@@ -133,6 +141,17 @@ Library "ethif-unix"
                       lwt,
                       lwt.unix
 
+Library "arpv4-unix"
+  CompiledObject:     best
+  Path:               unix
+  Findlibparent:      tcpip
+  Findlibname:        arpv4-unix
+  Modules:            Arpv4_unix
```
samoht: And I think you forgot that file to the PR :p

lib/wire_structs.ml
yomimono: OK, there's now an `arpv4_wire.ml` with the wire_structs defs (`arpv4` for consistency with other names), and an `Arpv4_wire` exposed alongside `Arpv4`.

mor1: In fact, pace previous discussions about pcap, shouldn't the wire formats
be placed into their own opam package for separate installation?

(ISTR someone saying that this could be done without splitting the repo via
opam.<package> or something...?)

On 6 May 2015 at 06:13, Thomas Gazagnaire <notifications@github.com> wrote:

> In lib/wire_structs.ml
> <https://github.com/mirage/mirage-tcpip/pull/134#discussion_r29756251>:
>
> > +    uint16_t ptype;
> > +    uint8_t hlen;
> > +    uint8_t plen;
> > +    uint16_t op;
> > +    uint8_t sha[6];
> > +    uint32_t spa;
> > +    uint8_t tha[6];
> > +    uint32_t tpa
> > +  } as big_endian
> > +
> > +  cenum op {
> > +    Op_request = 1;
> > +    Op_reply
> > +    } as uint16_t
> > +
> > +end
>
> I'm in favour of breaking the consistency of the library and put ARP wire
> format directly in arp.ml (or in arp_wire.ml).
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/mirage-tcpip/pull/134/files#r29756251>.
>



-- 
Richard Mortier
mort@cantab.net

samoht:
```
# Error: Unbound module type V1_LWT.ARP`
```

Can you add `PIN=mirage-types` to `.travis.yml` to pin the package to its dev version? Thanks!

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: c
 script: bash -ex .travis-ci.sh
 env:
-  - OCAML_VERSION=4.02 PACKAGE=tcpip MIRAGE_MODE=unix
-  - OCAML_VERSION=4.01 PACKAGE=tcpip MIRAGE_MODE=xen
+  - OCAML_VERSION=4.02 PACKAGE=tcpip MIRAGE_MODE=unix PINS=mirage-types:https://github.com/yomimono/mirage#separate_arp
+  - OCAML_VERSION=4.01 PACKAGE=tcpip MIRAGE_MODE=xen PINS=mirage-types:https://github.com/yomimono/mirage#separate_arp
```
samoht: the `url` should be `https://github.com/yomimono/mirage.git#separate_arp` (with `.git`) to make opam happy.

hannesm:
I really like this restructuring/separation of the arp layer.

samoht:
Seems to compile fine but the unit tests are failing ... not sure if this is related to these PR though.

MagnusS:
Hm it seems that the iperf test occasionally times out. I set the timeout to 30 seconds as the test completes in 4-5 seconds here. We could try increasing it to 60 seconds to see if it helps. Would it be possible to show the verbose test output in travis?

samoht:
the Travis machine can be very slow when running the tests, so yes it would be nice to have an adaptative timeout. To see the tests in travis you can use `--show-errors` in the opam file to be verbose on all failed errors.

MagnusS:
PR #137 should hopefully fix the iperf/travis test problem. The timeout is increased to two minutes and only half the data is transferred.

yomimono:
Thanks, @MagnusS !  We'll see whether that works a bit better.

yomimono:
I'm going to close this out, clean up the history, and make a new PR for the changes.

PR #138, opened 2015-05-19
Add IP allocation as a boot option

koleini:
A modification of ```tcpip``` so IP parameters can be passed via boot parameters:

```xl create network.xl 'extra="ip=192.168.1.100 gw=192.168.1.1 netmask=255.255.255.0"'```

So far, it works for unikernels with a single interface. To make it work for vms with multiple network devices, we have to access interface ids (or their string representations) and use command-line keys similar to ```tap0_ip```, etc. Accessing interface ids is not possible in ```netif``` (function ```id``` exists in ```netif.ml``` but not exposed in ```.mli```), and such function is also required for some other Mirage applications--for instance an OpenFlow switch needs to send interface ids to the controller.


samoht:
I agree there is a need for such feature but this should be done in the mirage tool itself with configure/runtime parameters that @Drup is currently working on. I'm closing this PR for now on because I expect the other patch to land soon. If that's not the case, please feel free to re-open and shout.

koleini:
I think having function ```id``` (or ```name``` ) back to the netif interfaces is required. Was there any reason for removing it?

samoht:
I think it's fine to export a string id for every netif implementation. /cc @MagnusS who knows well at least one other implementation ([vnetif](https://github.com/MagnusS/mirage-vnetif)). It has been removed at the same time we removed the `connect` functions from the device interface but we can add it back if needed.

I'm not sure to fully understand why you need to use that `id` function though. Could you try to elaborate a bit more?

talex5:
@koleini what do you want to use it for? If you know the concrete type (e.g. `mirage-net-xen`'s `Netif`) then the function is still available, but it's not clear what you'd do with a fully abstract one.

koleini:
Right, fully abstract one doesn't help. We need an interface that returns the name of the interface to the application. I have mentioned the reasons in my first comment.

samoht:
@koleini well, I've read:

> To make it work for vms with multiple network devices, we have to access interface ids (or their string representations) and use command-line keys similar to tap0_ip

but it is not totally clear to me why. That would be helpful if you elaborate more so that we can understand what is missing or to help you work-around the problem.

MagnusS:
In `vnetif` I assign an id to each netif when it's connected to a backend. The id is only needed by the backend and internally in vnetif (to pass to the backend), so it doesn't need to be exposed. I guess the generated mac can be used as an id though.

@koleini is the problem that you need to have netif's that map to a specific interface in dom0 to be able to assign the correct IP address? e.g. vifXX.XX in Xen or tap0 in Unix?

koleini:
@MagnusS exactly.

MagnusS:
Could you use the mac address to identify them?

PR #139, opened 2015-05-21
Remove uses of Io_page.

samoht:
Netif is now supposed to deal with page alignment.

### DIFF
--- a/channel/channel.ml
+++ b/channel/channel.ml
@@ -146,7 +146,7 @@ module Make(Flow:V1_LWT.FLOW) = struct
   (* Output functions *)
 
   let alloc_obuf t =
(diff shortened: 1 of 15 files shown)


avsm:
What's the rationale for this change? Won't this force a copy in Netif forever?

samoht:
Just a proposal after discussions with @hannesm and @pqwy as they thought the weird memory behaviour could maybe come from that part. Didn't change much things in practice though.

avsm:
I think this is a bad direction to go down, since it bakes in a copy where one doesn't need to exist. By being explicit about page alignment with pages that are allocated for write, we can do zero-copy write (even if the backend in Netif currently doesnt). With this patch, all that is lost...and I'm not sure what is gained.

samoht:
ok closing.

PR #141, opened 2015-05-25
Support for recording pcap files in mirage-vnetif tests

MagnusS:
Adds support for recording network traffic to pcap files in tests by adding a new listener to the mirage-vnetif backend. Currently enabled for the iperf and connect tests.


MagnusS:
The pcap files are currently stored where `test.byte` is executed from - usually in the root directory. Not sure if they should be stored somewhere else.

samoht:
We can maybe store them in `tests/pcap`? Would be useful to add there all the files needed to test for regressions.

MagnusS:
Yes, that's probably better. Is there a good way to find the relative path independently of where `test.byte` is executed from? Or can we assume that the tests are executed from the base directory?

samoht:
It's specified in the oasis file

MagnusS:
Ok, then I guess we can assume that the tests are always executed in the same working directory with `make test`. With the latest commit pcap files stored in `tests/pcap/*.pcap`.

samoht:
You need https://github.com/samoht/mirage-tcpip/commit/e301e27838969cdeb742ed7423b25b1b6fbd2b08 to fix the issue with Travis CI

MagnusS:
I've also cleaned up the vnetif tests and fixed a bug that broke the trailing_bytes tests. They just ran over the normal backend twice. The trailing_bytes tests are disabled for now though as they fail until #145 is merged.

MagnusS:
Looks like one of the travis test failed due to a network error this time

samoht:
I've merged #145 and re-run the failing tests.

samoht:
Tests are passing again. Magnus can you enable the trailing_bytes test again?

MagnusS:
Merged with master and enabled the tests - travis tests still pass

samoht:
Thanks!

PR #147, opened 2015-06-05
Add vnetif backend with uniform packet loss and new iperf test

MagnusS:
Adds a new vnetif backend + iperf test that drops 1% of sent packets


avsm:
very useful!

MagnusS:
Hm! One of the travis tests failed again - not sure why. I've enabled verbose test output in `opam` with -v. It would be better to use --show-errors, but it's not available in ocaml < 4.02

MagnusS:
Travis tests are passing now, so I guess it must have been a temporary problem.

samoht:
Nice!

PR #148, opened 2015-06-08
first cut at full RST processing for established connections

balrajsingh:


tcp/user_buffer.ml
```diff
@@ -317,4 +321,15 @@ module Tx(Time:V1_LWT.TIME)(Clock:V1.CLOCK) = struct
     clear_buffer t >>= fun () ->
     inform_app t
 
+  let rec dump_buffer t = 
+    match Lwt_sequence.is_empty t.buffer with
```
avsm: I think this could be shortened to

```
let dump_buffer t =
  Lwt_sequence.iter_node_l Lwt_sequence.remove t
```

(untested though)

samoht:
Thanks, merging! I'll clean-up some minor styling issues.

tcp/segment.ml
```diff
@@ -104,7 +105,20 @@ module Rx(Time:V1_LWT.TIME) = struct
   let input (q:t) seg =
     (* Check that the segment fits into the valid receive window *)
     let force_ack = ref false in
-    if not (Window.valid q.wnd seg.sequence) then return_unit
+    (* TODO check that this test for a valid RST is valid *)
+    if (seg.rst && (Window.valid q.wnd seg.sequence)) then begin
```
samoht: I'm not sure about that check, this seems to duplicate what we do just after that, eg. we could re-order the tests to:
1. test if the sequence is valid, if not `return_unit`
2. test if if it a `rst` (and we know that the sequence is valid)
3. do the rest

I'll propose a patch doing this.

PR #149, opened 2015-06-09
Iperf and vnetif improvements

MagnusS:
- Remove unnecessary calls to sleep
- Remove calls to flush when writing to pcap output channel
- Catch pcap write errors


MagnusS:
The listener function that writes to the pcap file is called with `Lwt.async` by the backend, resulting in a race when closing the pcap output channel. If packets are written to the channel while it is being closed this triggers a `Channel_closed` exception and the test fails. There is still a race with this patch, but the exception is caught. I have also removed the unnecessary call to `Lwt_io.flush` which should make it more likely that the recorder has completed before the channel is closed.

samoht:
Thanks!

PR #150, opened 2015-06-09
make Pcb.write fail if the connection is reset

samoht:


tcp/flow.ml
```diff
@@ -35,57 +40,54 @@ module Make(IP:V1_LWT.IP)(TM:V1_LWT.TIME)(C:V1.CLOCK)(R:V1.RANDOM) = struct
     | `Refused
   ]
 
+  let err_timeout () =
+    (* Printf.printf "Failed to connect to %s:%d\n%!" *)
+    (* (Ipaddr.V4.to_string daddr) dport; *)
```
avsm: Can't this use the `Log` module?

tcp/flow.ml
```diff
@@ -14,7 +14,12 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-open Lwt
+let (>>=) = Lwt.(>>=)
```
avsm: I think you can `open Lwt.Infix` to get these operators.

tcp/flow.ml
samoht: the log module is not merged yet, but yes that's the plan :-)

tcp/flow.ml
samoht: I never know in which version of `Lwt` the `Infix` module appeared. Thanks for the suggestion, I'll use that from now on.

PR #151, opened 2015-06-10
Prepare the release

samoht:


samoht:
I've tested mirage-seal on xen with the release and I confirm that the memory leak has completely disappeared. Well done everyone!!

PR #152, opened 2015-06-16
Use mirage-profile to implement Tcp.Log

talex5:
- Whenever we log a message to the console, also record it in the trace (if any). This means that all TCP log messages also show up on their associated thread in the trace viewer, and can be searched for.

- Counters are now implemented using `MProf.Counter`, which means they appear as a graph overlaid on the trace.

I made a slight change to the `Stats` API: I replaced `create` with `singleton`, since that seemed to reflect better what it was doing (if you did want multiple counters, MProf supports that too, but these counts look global to me anyway).

To see the log messages, you still need to enable logging. e.g. I used:

    let () =
      let open Tcp in
      [Pcb.info; Pcb.debug; State.debug] |> List.iter (fun log ->
        Log.enable log;
        Log.set_stats log false
      )

(turning off stats just prevents the numbers appearing in the messages, which is unnecessary since changes to the counters are logged anyway)

### DIFF
--- a/tcp/log.ml
+++ b/tcp/log.ml
@@ -25,12 +25,17 @@ type t = {
 
 let c = ref 0
 
+let write pp =
+  let msg = Format.flush_str_formatter () in
+  print_endline msg;
+  MProf.Trace.label msg
+
 let f t =
   if t.enabled && t.stats then
     let stats = Stats.create () in
-    fun pp -> Format.printf ("Tcp.%s%a: %t\n%!") t.name Stats.pp stats pp
+    fun pp -> Format.kfprintf write Format.str_formatter ("Tcp.%s%a: %t") t.name Stats.pp stats pp
   else if t.enabled then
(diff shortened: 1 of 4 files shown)


talex5:
(one Travis test failed with `Failure: iperf test timed out after 120.000000 seconds` - unlikely to be due to this patch, so I've restarted it)

samoht:
That seems nice!

talex5:
Note: this uses `Format.str_formatter` to buffer each message. Should create a new formatter just for logging?

samoht:
I don't know, need to profile that a little bit (running `./test.byte iperf 3` at the root of the repo is usually useful for that).

I think we should be able to create new logger output easily (if one wants to register a TCP logger for instance, that should only be a matter of `Log.register_logger <x>`, with a way to enable/disable loggers easily).

talex5:
I assume the plan is to move to a shared logging library at some point anyway?

samoht:
Thanks! I have some new ideas to speed-up the format strings

Issue #153, opened 2015-06-16
Invalid_argument("Lwt.wakeup_result") when closing a connection


talex5:
I have a simple TCP test client that writes a single byte and then closes the connection:

    match_lwt S.TCPV4.create_connection t (laptop_ip, 7001) with
    | `Error _err -> failwith "TCP error 7001"
    | `Ok flow ->
    
    let payload = Cstruct.sub buffer 0 1 in
    Cstruct.set_char payload 0 '!';

    match_lwt S.TCPV4.write flow payload with
    | `Error _ | `Eof -> assert false
    | `Ok () ->

    S.TCPV4.close flow >>= fun () ->
    OS.Time.sleep 1.0

Looking at a trace of this, I see some failed threads showing `Invalid_argument("Lwt.wakeup_result")`. It looks like `rx_application_t` tries to do `Lwt.wakeup urx_close_u ()` twice (the `data = None` case happens twice).

Not sure if it's a problem, but I guess that isn't supposed to happen...

![lwt-wakeup-fail](https://cloud.githubusercontent.com/assets/554131/8181312/c420d29a-141c-11e5-832a-f84abdc4fe9e.png)

talex5:
@yomimono / @balrajsingh any thoughts on how to fix this error?

    Fatal error: exception Invalid_argument("Lwt.wakeup_result")

What is the correct behaviour when closing a connection?
The error is coming from here:

https://github.com/mirage/mirage-tcpip/blob/455263dcc47409fbb1266c8a6553230a8b6f6d87/tcp/pcb.ml#L211

It looks like it sets `urx_close_u` and then calls itself recursively, which then tries to do it again.

To reproduce the bug, just apply the patch in #154 (which stops the error being silently ignored) and run the unit tests. You can see the error in the [Travis log](https://travis-ci.org/mirage/mirage-tcpip/jobs/67199445).

yomimono:
@talex5 - I'm looking at this now; I need to do a bit of puzzling over it before I have any suggestions.

yomimono:
It might be a good idea to page @MagnusS as well, as I know he was looking into mvar use in this chunk of code a while back. At first glance I'm not sure why we have `urx_close_u` and `urx_close_t` at all.

balrajsingh:
Doesn't look great, does it.  I too need to look at it more carefully and
think it about again.

The urx_close_u and urx_close_t was probably for an out-of-band close, ie
[...]

PR #154, opened 2015-06-17
RFC: If a TCP thread fails, call Lwt.async_exception_hook

talex5:
By default this will cause the unikernel to exit if the TCP code throws an exception. The application can override this by defining its own handler.

I think this is probably the correct behaviour, although it might break things due to pre-existing bugs (e.g. #153 - don't merge before fixing that!). At least, it should help us find other bugs!

### DIFF
--- a/tcp/pcb.ml
+++ b/tcp/pcb.ml
@@ -351,6 +351,10 @@ struct
     let fnth = fun _ -> th_frees := !th_frees + 1 in
     Gc.finalise fnpcb pcb;
     Gc.finalise fnth th;
+    Lwt.on_failure th (function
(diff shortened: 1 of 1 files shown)


yomimono:
I'm for it.

samoht:
Me too! Is someone looking at the other issue already?

yomimono:
@samoht - we're mulling it over in https://github.com/mirage/mirage-tcpip/pull/156 , which if enough people explain how Lwt works to me (using small words) might someday do the right thing.

talex5:
A better fix was applied in #156.

PR #155, opened 2015-06-19
Pull arpv4 module out of ipv4.

yomimono:
This is a rebase of https://github.com/mirage/mirage-tcpip/pull/134 , which pulls the arpv4 code out of the ipv4 module and provides independent access to arpv4 functions and an arpv4 record.

There are also some tests for the arp module included.  To run even slow-moving tests (such as cache expiry and timeouts for unresolvable addresses), run the tests without the -q option.

This PR shouldn't be merged without a matching PR for mirage, like the one at https://github.com/mirage/mirage/pull/419 .


talex5:
There are already problems in master with tests timing out on Travis. Do you think it would be possible to use a test clock/time rather than the real ones? That would also avoid the need for slow tests (unless actually CPU bound).

For example, here's the test clock I use for CueKeeper's unit tests (not quite the Mirage sig, though):

https://github.com/talex5/cuekeeper/blob/90a12e71834ae10416e0ec86ce15408ec25d33e6/test.ml#L24

yomimono:
Yes, I have a TODO about that in the comments for the test.  Thanks for 
the link; I wasn't aware you'd already done this in CueKeeper!

Thinking about it a bit more, for two of three slow tests in Arp it 
would be sufficient to just provide a Time module that divided the 
requested sleep time by some large factor; for the last (expiry) I think 
similarly simple tricks would work.

yomimono:
Arp tests now run in a parallel universe where time progresses at sixty times the normal rate.  I've reclassified them all as Quick for alcotest purposes, since they're pretty snappy now.  Thanks, @talex5 !

samoht:
Can you set `UPDATE_GCC_BINUTILS=1` in `.travis.yml` for the run testing Xen? this is needed for building xen-entropy.

Issue #157, opened 2015-06-27
Possible regression in 2.5.0 - iperf test w/packet loss frequently times out


MagnusS:
A few weeks ago I was able to run the iperf test w/uniform packet loss 150 times locally without timeout (as mentioned [here](https://github.com/mirage/mirage-tcpip/pull/142)) with the master branch. I just repeated the experiment with current master (f31810cb9a0a02745fa9dc6abb8162ee05e96c0a) and after a few attempts I've been unable to run more than max 8 tests in a row. Release 2.5.0 (455263dcc47409fbb1266c8a6553230a8b6f6d87) also times out frequently. 

With rev aab57095018b6afa8b6c30abc0e3c4957fe7fcea the test runs fine. To rule out a bug in the newest version of the test I've also tried using the tests from master with aab57095018b6afa8b6c30abc0e3c4957fe7fcea and got the same result (no timeouts). I also tried to double the timeout, in case it was caused by the reduced performance of the recently merged debug branch, but the test still times out - the connection seems to just stall when the test fails (see pcap output below).

This is the command I use - it runs 100 iperf tests and terminates if one of them fails. 
```
for c in $(seq 1 100); do echo $c; date; ./test.byte test iperf 2 -v || break ; done
```

I've also increased the data size from 10mb back to 25mb in [lib_tests/lib_iperf.ml](https://github.com/mirage/mirage-tcpip/blob/master/lib_test/test_iperf.ml#L220), as this was set lower to reduce timeouts in Travis. With 10mb the test is less likely to timeout, but it is still unreliable (I was able to run 20 tests with 10mb vs 8 with 25mb).

Here are the last packets in the pcap output from three failed tests.

```
28749 109.963717   10.0.0.101 -> 10.0.0.100   TCP 1514 [TCP Previous segment not captured] 14476 > 5001 [PSH, ACK] Seq=21089701 Ack=1 Win=262140 Len=1460
28750 109.963846   10.0.0.101 -> 10.0.0.100   TCP 1514 14476 > 5001 [PSH, ACK] Seq=21091161 Ack=1 Win=262140 Len=1460
28751 109.963906   10.0.0.100 -> 10.0.0.101   TCP 54 [TCP Dup ACK 28748#1] 5001 > 14476 [ACK] Seq=1 Ack=21088241 Win=262140 Len=0
[...]

PR #161, opened 2015-07-07
Prepare the release

samoht:
@talex5 @yomimono and @MagnusS can you check that you are happy with the change notes?


talex5:
OK with me (though it's not exactly a "dead-lock" - it's just very slow).

hannesm:
splitting out channel into a separate repository is not worth mentioning in changes?

yomimono:
Text looks OK but we should remove the channel tests (and surely the dependency?) as well.

MagnusS:
@MagnusS not @MagnusM :-) and Balraj also helped on #157

yomimono:
yaaaaaay! \o/

PR #164, opened 2015-07-29
ARP now handles ARP frames, not Ethernet frames with ARP payload

hannesm:
ARP until now received full ethernet frames, and defined ARP packets
in a way such that they include ethernet frames. This prevents code
reuse and is a discrepancy of the standard.

Furthermore, the data of the Ethernet header was never touched, and
the fix for assembling an ARP frame is trivial.


samoht:
I'll wait for Travis to be green and I'll add this to 2.6.0 ...

samoht:
The ARP tests don't compile.

hannesm:
I committed a fix just before you mentioned that it is broken :)

hannesm:
now travis is happy...

PR #166, opened 2015-08-10
Expose Ipv4.Routing.No_route_to_destination_address.

yomimono:
Calls to `Ipv4.write` and `Ipv4.writev` can fail if a link-level address for the destination IP can't be found via ARP; expose the exception generated in that case in `ipv4.mli` so users can match on it.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -8,8 +8,6 @@ bash -ex .travis-opam.sh
 export OPAMYES=1
 eval `opam config env`
 
-opam repo add mirage-dev https://github.com/mirage/mirage-dev.git
-
 git clone git://github.com/mirage/mirage-www
 cd mirage-www
 
--- a/lib/ipv4.mli
+++ b/lib/ipv4.mli
@@ -15,6 +15,11 @@
  *)
 
 module Make (N:V1_LWT.ETHIF) (A: V1_LWT.ARP) : sig
+  module Routing : sig
+    (* this exception can be thrown by `write` or `writev` when the destination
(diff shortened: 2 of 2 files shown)


yomimono:
Travis tests are failing because [mirage-www](https://github.com/mirage/mirage-www) can't be built; this seems to be because mirage-www is expecting a version of mirage with `archive_of_files`.  It looks like the mirage version installed is 2.6.0 which should include this function.  Any ideas?

samoht:
> opam repo add mirage-dev https://github.com/mirage/mirage-dev.git

this is adding my fork of `mirage/mirage`, which contains a hack to make mirage-www boots. I'll update my fork (which should be removed once https://github.com/mirage/mirage/pull/440 is released.

samoht:
This Travis logs are a bit hard to read ...

yomimono:
Not sure I understand, so let me summarize what I got out of that: because .travis-ci.sh is adding the mirage-dev repo, it's pulling your fork of `mirage/mirage`, which is versioned at 2.6.0 but missing the required patches for building storage from archives.  You're about to update your fork, and once that's complete re-running the tests should Just Work.  Correct?

samoht:
Yes that's correct. I've just update my fork so that should be better now. 

But after more thoughts, I think the correct fix would be to remove `opam repo add mirage-dev https://github.com/mirage/mirage-dev.git` from .travis-ci.sh

yomimono:
I've removed the offending line from `.travis-ci.sh` and the tests look better (yay!).

PR #167, opened 2015-08-14
wrap ports for listeners at 65536

yomimono:
Previously, it was possible to register listeners for ports > 65535.
This had unexpected results, as serializing the requested port number
elsewhere in the stack would result in truncating the port number, but
reply packets would not match any known listener (since the registered
listener was at the requested port number).  Register the given port
number mod 65536 to avoid this even if the user submits an unreasonable
port number, for consistency with other parts of the stack.

Discovered via https://github.com/mirage/ocaml-dns/blob/master/mirage/dns_resolver_mirage.ml#L108 .

### DIFF
--- a/lib/tcpip_stack_direct.ml
+++ b/lib/tcpip_stack_direct.ml
@@ -77,10 +77,10 @@ struct
   let ipv4 { ipv4; _ } = ipv4
 
   let listen_udpv4 t ~port callback =
-    Hashtbl.replace t.udpv4_listeners (port mod 65536) callback
+    Hashtbl.replace t.udpv4_listeners port callback
 
   let listen_tcpv4 t ~port callback =
-    Hashtbl.replace t.tcpv4_listeners (port mod 65536) callback
+    Hashtbl.replace t.tcpv4_listeners port callback
 
   let configure_dhcp t info =
     Ipv4.set_ip t.ipv4 info.Dhcp.ip_addr
--- a/lib/tcpip_stack_direct.ml
+++ b/lib/tcpip_stack_direct.ml
@@ -63,8 +63,8 @@ struct
     ipv4  : Ipv4.t;
     udpv4 : Udpv4.t;
     tcpv4 : Tcpv4.t;
-    udpv4_listeners: (int, Udpv4.callback) Hashtbl.t;
-    tcpv4_listeners: (int, (Tcpv4.flow -> unit Lwt.t)) Hashtbl.t;
+    udpv4_listeners: (Cstruct.uint8, Udpv4.callback) Hashtbl.t;
+    tcpv4_listeners: (Cstruct.uint8, (Tcpv4.flow -> unit Lwt.t)) Hashtbl.t;
   }
 
   type error = [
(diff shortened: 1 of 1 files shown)


lib/tcpip_stack_direct.ml
```diff
@@ -77,10 +77,10 @@ struct
   let ipv4 { ipv4; _ } = ipv4
 
   let listen_udpv4 t ~port callback =
-    Hashtbl.replace t.udpv4_listeners port callback
+    Hashtbl.replace t.udpv4_listeners (port mod 65536) callback
```
nojb: Probably using `abs port` instead of `port` would be better to avoid negative port numbers.  And probably when you get zero, the port should be chosen randomly among the set of free ports.

hannesm: I think using `mod 65536` is counterintuitive here: if a developer calls `listen_xxx` with a `port` greater than `2^16`, there's no way to find out on which port the service actually listens.

instead, I'd document the 16 bit size of port clearly and throw an exception if violated.

yomimono:
@nojb - I think the user often just wants to choose randomly among the set of free ports and this should be more explicitly supported in the API, not just implied by the absence of an argument.  The user that led me to discover this behavior, for example, wants both to send a packet and establish a listener, and so in order to send with the correct source port will need to know which port the callback is listening on.

@hannesm - in that case I'll probably add an exception for the port already being in use (in TCP as well), since this is another error case that currently generates unexpected results.  I'd prefer to change the API generally to return a result type rather than throw an exception, but that discussion is edging on territory in https://github.com/mirage/mirage/issues/426 .

lib/tcpip_stack_direct.ml
avsm: Agreed, let's go for forbidding incorrect ports rather than the wrapping behaviour.

samoht:
I think we haven't reach agreement on this PR yet, so I'm not merging it. Seeing from far away, I find the implicit modulo behaviour a bit counter-intuitive, though.

mor1:
A bit late to the party but my tuppence: 

1. Port numbers have a well-defined range and should be expected to be within that range-- users trying to bind to ports outside that range should get an error of some kind. The wrapping behaviour here seems very odd and likely to lead to much confusion.

2. It seems within the "generally be explicit" approach Mirage mostly takes to provide an API that returns a free (ephemeral) port number for the caller to bind to later if they wish; but an approach that looks more like traditional sockets may be fine too. (The former might require some work to map onto Unix sockets backend.)

matildah:
9450089e59b37bd6e953a1294e2f75d92a492287 contains a patch that raises an exception if listen_udpv4 or listen_tcpv4 is given a port number that isn't between 0 and 65535.

hannesm:
@matildah great. can you submit a pull request to this repository with the commit? thx!

mor1:
The update to the signatures of `udp4_listeners` and `tcp4_listeners` seems useful, but not the port number wrapping.

@yomimono @hannesm @matildah Best way to handle? Either @yomimono updates this PR I guess, or just close this one and @matildah adds the signature change to #173 ??

yomimono:
Updated to remove the wrapping logic; sorry for my lack of participation in this discussion of late.

hannesm:
While the idea, to use `Cstruct.uint8` here, is good; it is only partial and does not improve much: [Cstruct.uint8](https://github.com/mirage/ocaml-cstruct/blob/master/lib/cstruct.mli#L149) is an alias for `int`, and users of these hash tables pretend it to be integers -- thus it will break once (if ever) Cstruct decides on another representation for `uint8`...

mor1:
Hm. Ok, in light of comments on #173, I'm inclined to just close this for now, merge #173 and then await many, no doubt imminent, PRs with the fixes described therein...

@hannesm @yomimono @samoht Ok?

yomimono:
Fine by me.

PR #170, opened 2015-08-26
Add optional arguments for settings in ip v6 and v4 connects

Drup:
This simplifies the code a bit in the mirage dsl and it should be useful in general.


avsm:
Fine by me, but could you add ocamldoc to indicate what the default values are for that function?

Drup:
I can do it for ipv4, but for ipv6, I have no idea what the defaults are. When an argument is not present, I just don't set it.

Drup:
Should I add a default value ? I must admit that I basically don't understand the rest of the file.

samoht:
thanks!

PR #171, opened 2015-09-01
only use `opam config var prefix` if dependencies are not present

gfxmonk:
This allows building without the `opam` command line tool installed as long as pkgconfig dependencies are already available (as is the case with `opam2nix`).

### DIFF
--- a/postconf.ml
+++ b/postconf.ml
@@ -28,7 +28,8 @@ let () =
 
   let xen_cflags =
     if !xen then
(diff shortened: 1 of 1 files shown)


samoht:
I'm re-running the failed Travis build.

Can you replace the catch-all pattern (`_`) but a more precise exception? This script is already crazy enough, would be better to make it more robust if possible.

samoht:
Travis is green, merging. I'll add a less generic catch-all handler before releasing.

gfxmonk:
Thanks!

PR #173, opened 2015-10-11
raise Failure exception if given an invalid port number in listen_udpv4 or listen_tcpv4

matildah:


hannesm:
looks good to me (supersedes #167)

mor1:
Agree. AIUI we mostly want to move away from exceptions at some point, but this seems strictly better to me than what's currently there.
Travis is green-- @samoht any objections to merge?

hannesm:
@mor1 moving away from exceptions is the ultimate goal, but will take some time. this is a good quickfix for now.

dbuenzli:
It seems people are still a little bit confused about how to do proper error handling in OCaml. In this case either:

1. The function should check the range in which case you want to return a `result` type with an error if the client provides a wrong number.
2. The client should provide an integer in the appropriate range, in which case this should be properly documented in the interface of the function and `Invalid_argument` should be raised (denoting a programming error) if the client provides a wrong number.

In general if you find yourself using the `Failure` exception you can safely assume that you are doing something wrong.

talex5:
I think "proper" error handling in this case would be to change the type from `int` to `port`, so you can only pass in a valid port number. Perhaps `mirage-types` could provide a suitable type (e.g. `type port = private int`)? We should only need to check a port is valid when parsing external input, not in every function that takes a port (to produce the error) and every function that calls such a function (to handle the error).

dbuenzli:
@talex5 that is clearly the best to do, I was commenting on the specific commit.

hannesm:
first of all, I proposed @matildah to work on this issue and solution as a first contribution (in the terms of [the gnome outreachy program](https://wiki.gnome.org/Outreachy)); thus please blame me.

I overall prefer the solution to explicitly return success/failure (or use a `port` type), but unfortunately both solution will break the API (and I don't think it is worth it).

@dbuenzli you're right that `invalid_arg` is a better exception here.  Additionally, the bounds checks should be applied to the sockets implementation ([udp](https://github.com/mirage/mirage-tcpip/blob/master/unix/tcpip_stack_socket.ml#L74) and [tcp](https://github.com/mirage/mirage-tcpip/blob/master/unix/tcpip_stack_socket.ml#L96)), and documented in [mirage-types](https://github.com/mirage/mirage/blob/master/types/V1.mli#L714-L724).

dbuenzli:
> thus please blame me.

No one is being blamed here. We are just trying to make sure that the best things are being done. 

> I overall prefer the solution to explicitly return success/failure (or use a port type)

If the API is ever broken returning a `result` type would clearly be inferior in that case. Having a `port` type allows for a stronger invariant that can be relied upon over the complete code base.

lib/tcpip_stack_direct.ml
```diff
@@ -77,10 +77,15 @@ struct
   let ipv4 { ipv4; _ } = ipv4
 
   let listen_udpv4 t ~port callback =
-    Hashtbl.replace t.udpv4_listeners port callback
+    if port < 0 || port > 65535
+    then raise (Failure "invalid port number")
```
dbuenzli: The error message should be factored out into something like this
```ocaml
let err_invalid_port p = Printf.sprintf "invalid port number (%d)" p
```
And the raised exception should be `Invalid_argument`.

mor1:
Ok, I propose the following:

@matildah Please could you factor out the error message into a function that returns the message to be raised, and then raise `Invalid_argument` rather than `Failure`, per @dbuenzli suggestions, and update this PR. I will then merge it.

@hannesm @dbuenzli @talex5 I think with the small changes suggested above this is clearly worth merging as it strictly improves on what is currently done. More thorough and correct refactoring can then be undertaken by anyone who cares to make a start on what I suspect will be a fairly pervasive set of API breaking changes.

dbuenzli:
@mor1 Documentation must also be updated before merging.

dbuenzli:
> @mor1 Documentation must also be updated before merging.

Ah well it's in another repo... But it would be nice if @matildah could update the documentation there.

mor1:
Sure-- @matildah if you could submit another PR to [the docs](https://github.com/mirage/mirage/blob/master/types/V1.mli#L714-L724) to note that the `port:int` parameter is bounds checked to ensure it's inside the valid rnage for ports, that'd be great :)

matildah:
I changed the pull request to raise `Invalid_argument` and also did https://github.com/mirage/mirage/pull/449 for the docs

samoht:
Thanks!

PR #174, opened 2015-10-13
improve tcp options marshalling/unmarshalling

yomimono:
This PR includes some tests for TCP options marshalling and unmarshalling, as well as some changes to make them pass.  Among cases addressed are the problems in #106 .

PR #176, opened 2015-10-20
some state tests + fixes for closure conditions

yomimono:
PR is based on #174 rather than master, so the changeset is smaller than it initially appears.

Resolve a couple of TODOs:
* check ACK numbers when processing RSTs received after sending an initial SYN
* make `Tcp.State.t` abstract

In addition, a few more changes:
* replace the `Reset` state with `Closed`
* fix a bug where timers for `Fin_wait_2` would never resolve if multiple ACKs to a sent FIN were received
* process RSTs only when the sequence number exactly matches the left edge of the receive window (see [RFC 5961](https://tools.ietf.org/html/rfc5961))

And some added tests:
* a bunch of tests for `tcp/state.ml`.  Note that logic around checking seq/ack numbers is scattered between different modules in `tcp/`, so these tests can't be complete by targeting only this module.
* `Tcp.create_connection` reports failure immediately and correctly when the initial SYN receives an RST/ACK in response

### DIFF
--- a/lib_test/test_connect.ml
+++ b/lib_test/test_connect.ml
@@ -83,10 +83,39 @@ module Test_connect (B : Vnetif_backends.Backend) = struct
 
     Lwt.return_unit
 
+  let test_tcp_connect_to_closed_port () =
+    or_error "console" Console.connect "console" >>= fun c ->
+    let timeout = 15.0 in
+    Lwt.pick [
+      (Lwt_unix.sleep timeout >>= fun () ->
+       fail "connect test timed out after %f seconds" timeout) ;
+
+      (V.create_stack c backend server_ip netmask [gw] >>= fun s1 ->
+       (* register no listeners *)
+       V.Stackv4.listen s1);
+
+      (Lwt_unix.sleep 0.1 >>= fun () ->
+       V.create_stack c backend client_ip netmask [gw] >>= fun s2 ->
+       V.Stackv4.TCPV4.create_connection (V.Stackv4.tcpv4 s2) (server_ip, 80) >>= function
+       | `Ok _ -> fail "create_connection claimed success connecting to a stack with no listeners"
+       | `Error e -> match e with
+         | `Refused -> Lwt.return_unit (* yay! *)
+         | `Timeout -> fail "create_connection got timeout when it should've gotten rst"
+         | `Unknown s -> fail "wrong kind of error in create_connection"
+      )
+    ] >>= fun () ->
+    Lwt.return_unit
+
   let record_pcap =
     V.record_pcap backend
 
 end
+
+let test_tcp_connect_to_closed_port_basic () =
+  let module Test = Test_connect(Vnetif_backends.Basic) in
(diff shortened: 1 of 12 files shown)


yomimono:
If someone could kick Travis, I'd appreciate it; 4.01 failed on a 404 downloading dependencies.

talex5:
Kicked!

yomimono:
Any thoughts on this?

lib_test/test_connect.ml
```diff
@@ -83,11 +83,40 @@ module Test_connect (B : Vnetif_backends.Backend) = struct
...
+    let timeout = 15.0 in
+    Lwt.pick [
+      (Lwt_unix.sleep timeout >>= fun () ->
+       fail "connect test timed out after %f seconds" timeout) ;
+
+      (V.create_stack c backend server_ip netmask [gw] >>= fun s1 ->
+       (* register no listeners *)
+       V.Stackv4.listen s1);
+
+      (Lwt_unix.sleep 0.1 >>= fun () ->
```
talex5: I dislike unit tests with real sleeps in them. Apart from being unreliable, when building lots of packages they all add up...
But still better than not having a test, so not a blocker for the PR.

tcp/state.mli
```diff
@@ -40,30 +41,33 @@ type tcpstate =
...
 val state : t -> tcpstate
-val t : on_close:close_cb -> t
+val start : on_close:close_cb -> t
 
 val pp: Format.formatter -> t -> unit
 
 module Make(Time : V1_LWT.TIME) : sig
   val fin_wait_2_time : float
+  (* when in state fin_wait_2, use this as a timeout parameter  *)
   val time_wait_time : float
```
talex5: Slightly ambiguous which value the comment refers to (`fin_wait_2_time` or `time_wait_time`).

tcp/state.ml
```diff
@@ -123,42 +119,45 @@ module Make(Time:V1_LWT.TIME) = struct
...
-          Lwt.async (fun () -> finwait2timer t count fin_wait_2_time);
-          Fin_wait_2 count
-        else
+        if diffone b a then begin
+          Lwt.async (fun () -> finwait2timer t fin_wait_2_time);
+          Fin_wait_2
+            (* so the first time we enter fin_wait_2, we set the count to 0 and
+               spawn a waiting thread; the waiting thread only closes the
+               connection if the state still matches the number.  So if we keep
+               receiving ACKs, won't we never close? That seems incorrect. *)
```
talex5: Out-of-date comment? Looks like you removed the counter.
@balrajsingh any idea what this counter scheme was for?

talex5:
I'm not too familiar with the TCP state model, but the parts of the patch I understand look good :-) Would like to understand what the fin counter stuff was for though - presumably it was added for a reason, even if it's wrong in other ways.

yomimono:
I'll separate out a few things and submit a new PR which doesn't conflict with the just-merged #182 .

PR #177, opened 2015-10-20
fix #162 (duplicate ACKs during closure)

### LINKED ISSUE #162: Duplicate ACKs sent when closing a connection
Can be seen in the pcap files produced by the tests. Examples:

```
$ tshark -nr tests/pcap/tcp_iperf_two_stacks_basic.pcap | tail -n 5
13710   1.694047   10.0.0.100 -> 10.0.0.101   TCP 54 5001 > 18442 [FIN, ACK] Seq=1 Ack=10000002 Win=262140 Len=0
13711   1.694124   10.0.0.101 -> 10.0.0.100   TCP 54 18442 > 5001 [ACK] Seq=10000002 Ack=2 Win=262140 Len=0
13712   1.694272   10.0.0.100 -> 10.0.0.101   TCP 54 [TCP Dup ACK 13710#1] 5001 > 18442 [ACK] Seq=2 Ack=10000002 Win=262140 Len=0
13713   1.694357   10.0.0.101 -> 10.0.0.100   TCP 54 [TCP Dup ACK 13711#1] 18442 > 5001 [ACK] Seq=10000002 Ack=2 Win=262140 Len=0
13714   1.694397   10.0.0.100 -> 10.0.0.101   TCP 54 5001 > 18442 [RST, ACK] Seq=2 Ack=10000002 Win=0 Len=0
```
```
$ tshark -nr tests/pcap/tcp_iperf_two_stacks_uniform_packet_loss.pcap|tail -n 5
13634  78.002340   10.0.0.100 -> 10.0.0.101   TCP 54 5001 > 19123 [FIN, ACK] Seq=1 Ack=10000002 Win=262140 Len=0
13635  78.002417   10.0.0.101 -> 10.0.0.100   TCP 54 19123 > 5001 [ACK] Seq=10000002 Ack=2 Win=262140 Len=0
13636  78.002566   10.0.0.100 -> 10.0.0.101   TCP 54 [TCP Dup ACK 13634#1] 5001 > 19123 [ACK] Seq=2 Ack=10000002 Win=262140 Len=0
13637  78.002652   10.0.0.101 -> 10.0.0.100   TCP 54 [TCP Dup ACK 13635#1] 19123 > 5001 [ACK] Seq=10000002 Ack=2 Win=262140 Len=0
13638  78.002698   10.0.0.100 -> 10.0.0.101   TCP 54 5001 > 19123 [RST, ACK] Seq=2 Ack=10000002 Win=0 Len=0
```

```
$ tshark -nr tests/pcap/tcp_connect_two_stacks_basic.pcap | tail -n 5
  8   0.102135   10.0.0.101 -> 10.0.0.100   TCP 91 [TCP segment of a reassembled PDU]
  9   0.102217   10.0.0.100 -> 10.0.0.101   TCP 54 80 > 15426 [ACK] Seq=1 Ack=38 Win=262140 Len=0
 10   0.102437   10.0.0.101 -> 10.0.0.100   TCP 54 15426 > 80 [FIN, ACK] Seq=38 Ack=1 Win=262140 Len=0
 11   0.102514   10.0.0.100 -> 10.0.0.101   TCP 54 80 > 15426 [ACK] Seq=1 Ack=39 Win=262140 Len=0
 12   0.102665   10.0.0.100 -> 10.0.0.101   TCP 54 [TCP Dup ACK 11#1] 80 > 15426 [ACK] Seq=1 Ack=39 Win=262140 Len=0
```

Related to this? (from @balrajsingh in #140)

>As you say there is a dup-ack near the beginning but that's from the
"server" and it's acking the FIN that was send from the client, after it
half-closed the connection.  This particular ack is redundant and shouldn't
be there, but it really is quite harmless and doesn't have any effect on
the connection.  I think it's because when a FIN is received an ACK is sent
back immediately, but the delayed-ACK timer thinks the ACK wasn't sent.
The cause of the confusion is probably because the seq number that was
advanced for ACK-ing the FIN was just done adhoc in that packet - the
delayed-ACK timer is looking at the overall connection's seq number.  It
should be fixed, but this one is harmless.

yomimono:
Signaling to the thread that "spools the data into an application
receive buffer" was incorrectly being treated equally between the situation
where we need to acknowledge data but have none to send ourselves, and
the situation where the connection is being closed.  This resulted in a
duplicate ACK being sent after any ACK to a FIN/ACK segment.  Instead,
check to see whether the connection is being closed; if it is, don't
cause another ACK to be sent early.

### DIFF
--- a/tcp/pcb.ml
+++ b/tcp/pcb.ml
@@ -192,23 +192,23 @@ struct
       (* Thread to monitor application receive and pass it up *)
       let rec rx_application_t () =
         Lwt_mvar.take rx_data >>= fun (data, winadv) ->
-        begin match winadv with
+        let signal_ack = function
           | None        -> Lwt.return_unit
-          | Some winadv ->
-            if (winadv > 0) then (
+          | Some winadv when winadv > 0 ->
               Window.rx_advance wnd winadv;
               ACK.receive ack (Window.rx_nxt wnd)
-            ) else (
+          | Some winadv ->
               Window.rx_advance wnd winadv;
               ACK.pushack ack (Window.rx_nxt wnd)
-            )
-        end >>= fun _ ->
+        in
         begin match data with
           | None ->
+            (* don't send an ACK in this case; this already happened *)
             STATE.tick pcb.state State.Recv_fin;
             User_buffer.Rx.add_r urx None >>= fun () ->
             Lwt.return_unit
           | Some data ->
+            signal_ack winadv >>= fun () ->
             let rec queue = function
               | []     -> Lwt.return_unit
               | hd::tl ->


talex5:
I don't really understand this code, but it sounds reasonable (and the old behaviour was certainly buggy).

I see that this code gets triggered with `None` from two places in segment.ml. In both cases, we send `(None, Some 0)`. Perhaps this type could be changed to make it clearer that the `Some 0` isn't used. e.g.

    type rx_data =
      | Data of Cstruct.t list * int option
      | Closing_connection

Why does the code go to the trouble of passing `Some 0` rather than `None` for `winadv`, which would also avoid the problem? Is there any case where the ack is still needed?

(git blame ends with `initial import of mirage networking libraries` unfortunately)

/cc @balrajsingh

balrajsingh:
This was a bit of signalling that had to be added onto an existing channels
- the result is admittedly ugly.  It should be redone cleaner.

The signalling is:

If data is None it means a FIN has been received, else a data segment.

If winadv is None it means an ACK does not need to be sent (nothing to be
ACK-ed).
If winadv is Some w where w > 0 then it means a segment was received
in-order and a normal ACK where ack-compression, delayed-acks and other
techniques can be applied.
If winadv is Some w where w is 0, then an ACK must be pushed out
immediately - these are duplicate ACKS for signalling out-of-order or lost
packets and every one of them must be sent.

Odd about blame not showing history correctly.

yomimono:
Merging at @talex5's urging.

PR #178, opened 2015-10-26
Remove bogus warning about ignored socket config

talex5:
If the client requests binding to V4.any then it reports:

    Manager: socket config currently ignored (TODO)

and then binds to it anyway. Skip the warning in this case.


avsm:
+1

PR #179, opened 2015-10-28
Ipv6 cleanup

nojb:
I was sitting on this cleanup for a long time without time to get it into shape.

There are no external changes to the lwt interface.

The logic is essentially unchanged but with a fair amount of code reorganisation, simplification, etc.  I think it will help to whoever wants to hack on this.  Also I added a proper signature with some comments.

### DIFF
--- a/lib/ipv6.ml
+++ b/lib/ipv6.ml
@@ -38,11 +38,15 @@ module Make (E : V1_LWT.ETHIF) (T : V1_LWT.TIME) (C : V1.CLOCK) = struct
 
   let id { ethif } = ethif
 
-  let tick t =
-    let now = C.time () in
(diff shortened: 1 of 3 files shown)


hannesm:
what is the status of this PR? should we merge it, considering that nobody apart from @nojb really used the v6 stuff (and according to him, this PR improves on readability).

nojb:
This PR can be merged as it is - it improves quite a bit the organization of the code and general readability so I don't see any downsides to merging ...

hannesm:
@djs55 @samoht @avsm @yomimono @talex5 anything against merging this PR?

avsm:
Fine with this -- we need to integrate IPv6 properly via Functoria's config interface, so the latest version in trunk is best.  No consumers of this yet as far as I know.

PR #182, opened 2016-01-22
Rst checking,  rfc5961

ppolv:
Hi, first time digging into mirage code,  I'm interested on learning more about the TCP/IP implementation.

Started with what seems an easy fix for rst on  syn_sen state:
https://github.com/mirage/mirage-tcpip/blob/185172167d5ccd86e2d725eab4aa68f1d760fa2c/tcp/pcb.ml#L399

but added also the mitigations to blind attacks described in rfc5961 

The change itself is simple.  Testing it is more challenging.   I *tried* (lib_test/test_rfc5961.ml
) to exercise the new code and make it easy to add new tests of this kind, with crafted tcp packets being injected.  Is there a better way already?,  mine feels  verbose and error prone.

(you could also try the tests on current master to see the ones that fail).

### DIFF
--- a/lib_test/common.ml
+++ b/lib_test/common.ml
@@ -7,6 +7,11 @@ let or_error name fn t =
   | `Error e -> fail "or_error starting %s" name
   | `Ok t    -> Lwt.return t
 
+let expect_error error name fn t =
+  fn t >>= function
+  | `Error error -> Lwt.return t
+  | _    -> fail "expected error on %s" name
+
 let assert_string msg a b =
   let cmp a b = String.compare a b = 0 in
   OUnit.assert_equal ~msg ~printer:(fun x -> x) ~cmp a b
--- a/lib_test/common.ml
+++ b/lib_test/common.ml
@@ -9,7 +9,8 @@ let or_error name fn t =
 
 let expect_error error name fn t =
   fn t >>= function
-  | `Error error -> Lwt.return t
+  | `Error error2 when error2 = error ->
+          Lwt.return t
   | _    -> fail "expected error on %s" name
 
 let assert_string msg a b =
--- a/lib_test/test.ml
+++ b/lib_test/test.ml
@@ -19,6 +19,7 @@ let suite = [
   "connect"        , Test_connect.suite     ;
   "iperf"          , Test_iperf.suite       ;
   "tcp_options"    , Test_tcp_options.suite ;
+  "rfc5961"        , Test_rfc5961.suite     ;
 ]
 
 let run test () =
--- a/lib_test/test_rfc5961.ml
+++ b/lib_test/test_rfc5961.ml
@@ -1,0 +1,395 @@
+(*
+ * Copyright (c) 2016 Pablo Polvorin <pablo.polvorin@gmail.com>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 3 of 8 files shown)


samoht:
Thanks for your contribution! And extra-bonus with the added tests, this looks really good.

/cc @balrajsingh @yomimono @hannesm and @MagnusS who are the network experts to review

hannesm:
awesome! looks related to #176 -- I will hopefully have time soon to look into both PRs (but unlikely before February)

hannesm:
also worth to point to this great informative article on RST handing in various operating systems https://www.snellman.net/blog/archive/2016-02-01-tcp-rst/

ppolv:
indeed it overlaps in part with #176, sorry I didn't realize there was work already done on this pending for review/merge,  my bad.

diff regarding this is one is here I send challenge acks back,  and the check done for blind data injection in syncronized state, also from a recommendation in rfc5961.

Comparing with the per-OS table of rst handling in https://www.snellman.net/blog/archive/2016-02-01-tcp-rst/,  the patch will be closer to linux behaviour:
(assumes a last ACK sent of 12, last sequence number received of 17, and a window of 10). 

10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 | 21 | 22 | 23 | 24 | 25 | 26 | 27 | 28 | 29
--- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- 
A | A | A | A | A |  A | A |  R | A | A | A | A  | A | D | D | D | D | D | D | D  

A=ACK, R=Reset, D=Drop

The difference is in the lower end.  If the reset has sequence 10,  it doesn't exactly match the rcv_next,  but is in the "validity window" as computed by mirage: 

```ocaml
let valid t seq =
  let redge = Sequence.(add t.rx_nxt (of_int32 t.rx_wnd)) in
  let ledge = Sequence.(sub t.rx_nxt (of_int32 t.max_rx_wnd)) in
  let r = Sequence.between seq ledge redge in
```
and so we send a challenge ack.


Linux consider it outside of the window  and just drop it without ack. http://lxr.free-electrons.com/source/net/ipv4/tcp_input.c#L5151

The check is more complete on linux/bsd case,  it checks if *any* part of the this segment (start, start + len + (fin|syn) ) falls inside the window. And the window is defined as starting at rcv_wup, for that 10 is outside it,  it is < than the last ack sent that was 12.   
Not sure the implications of this,  might this be on purpose to keep the code simpler?.

Reading the linux code for the window check, found other difference that could be problematic.  RFC indicates we should sent an ack in response to data outside window(unless it is a rst) and discard the incoming segment [RFC793 p.69]. But mirage discards it without sending ack (both on master and on this pull) https://github.com/mirage/mirage-tcpip/blob/185172167d5ccd86e2d725eab4aa68f1d760fa2c/tcp/segment.ml#L131
Linux sends an ack in that case http://lxr.free-electrons.com/source/net/ipv4/tcp_input.c#L5164 ,  

and if I'm reading it well,  so do openbsd:
http://fxr.watson.org/fxr/source/netinet/tcp_input.c?v=OPENBSD#L1360
http://fxr.watson.org/fxr/source/netinet/tcp_input.c?v=OPENBSD#L1409

Here *might* be problems too 
https://github.com/mirage/mirage-tcpip/blob/185172167d5ccd86e2d725eab4aa68f1d760fa2c/tcp/segment.ml#L148
mirage don't 'cut' data and keep only the one in window, but uses the entire segment instead.
If the data the other end sends is overlapping with some data we already have,  ej:
we receive  seq  10, with len 2. So rx_nxt is 13.  Then receive  seq 10, with len 100. Mirage think *all* that is in the past and already processed. I don't know under what conditions in current implementation could lead to a peer to send this kind of data, afaik mirage-tcpip always acks entire segments and not part of it. Don't know enough about it to be sure if the other end might or must not send overlappings like that.  
I guess there might be some corner cases this could happens, for example if our acks are lost,
and the other ends timeout and resend, but now coalescing with further data.. like sending seq 10, len 102 in the retry?.  Or if mirage at some point implements syn cookies and acks the syn but discard any initial data on it.

yomimono:
Thanks for your thoughtful comments, and the PR!  I particularly liked your test code which I found to be quite thorough, although it took me some time to understand.  I need a bit more time to review in more detail, but the patch is generally good IMO.

Good point about the behavior when we receive data outside the window; I agree that we should discard and send an ACK instead of silently discarding the packet.

Mirage generally is poorly equipped to handle overlapping segments at the moment - there are a few other TODOs extant related to this.  While this would be good to fix, I think it should be a separate PR.

yomimono:
Unless there are objections, I'd like to merge this soon.

PR #184, opened 2016-03-12
Improve some error strings in ipv4

yomimono:
The ipv4 module gives some misleading error output in a few cases, notably when the network is unreachable.  Change the strings to be less misleading and to include more information.

This patch doesn't do anything to make error output nicer (e.g. use the Log library), it just makes the strings a bit better.

### DIFF
--- a/lib/ipv4.ml
+++ b/lib/ipv4.ml
@@ -176,7 +176,7 @@ module Make(Ethif: V1_LWT.ETHIF) (Arpv4 : V1_LWT.ARP) = struct
       let frame = Cstruct.set_len frame header_len in
       write t frame buf
     |ty ->
-      printf "ICMP unknown ty %d\n" ty;
+      printf "ICMP unknown ty %d from %s\n" ty (Ipaddr.V4.to_string src);
       Lwt.return_unit
 
   let input t ~tcp ~udp ~default buf =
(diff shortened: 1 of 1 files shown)


djs55:
As someone who has encountered some of those error conditions recently, the new messages look better to me!

samoht:
That build error on 4.01 is weird:

```
make[1]: Leaving directory `/home/travis/build/mirage/mirage-tcpip/mirage-www/stats'
mirage configure -f src/config.ml  -t xen
mirage: unknown option `-t'.
``

**Comment by yomimono** [comment]
CI is failing because there's a `mirage configure -t xen` somewhere in the CI instructions, which no longer works as of mirage 2.7.0 .  I'll try to find it.

**Comment by avsm** [comment]
Constraints picking up older packages probably

File: lib/ipv4.ml [hunk 1]
```diff
@@ -133,35 +136,33 @@ module Make(Ethif: V1_LWT.ETHIF) (Arpv4 : V1_LWT.ARP) = struct
...
 
   let icmp_input t src _hdr buf =
     MProf.Trace.label "icmp_input";
     match Wire_structs.Ipv4_wire.get_icmpv4_ty buf with
     |0 -> (* echo reply *)
-      printf "ICMP: discarding echo reply\n%!";
+      printf "ICMP: discarding echo reply from %s\n%!" (Ipaddr.V4.to_string src);
       Lwt.return_unit
-    |3 -> icmp_dst_unreachable buf
+    |3 -> printf "icmp_dst_unreachable buf"; Lwt.return_unit
```
talex5: Typo? Presumably we want to call `icmp_dst_unreachable buf`, not print it.

PR #192, opened 2016-03-20
Prepare to release 2.7.0

djs55:
This is part of the great cstruct.syntax purge -- see [mirage/ocaml-cstruct#95]

Many other changes are also included!


hannesm:
would be great to have this merged + tcpip without camlp4 released! :)

PR #193, opened 2016-03-23
Pull ICMP functionality into its own module

yomimono:
Remove the implicit ICMP behavior from IPV4 and instead provide an ICMPV4 implementation.  Include some simple tests for this module, and provide sensible default behavior for the STACKV4 modules (in other words, answer pings).  While this PR includes little new functionality, it should make developing ICMP-related features easier for consumers of the stack (e.g. mirage-nat, mirage-mimic, and other router- or proxy-like applications).

An incidental bugfix for the indiscriminate answering of ICMP echo requests is also included.


hannesm:
this is great!

it should now be safe to remove the ICMP code from ipv4/ipv6 (eg https://github.com/yomimono/mirage-tcpip/blob/icmp_refactor/lib/ipv4.ml#L138)!?

yomimono:
Yes for ipv4 (thought I'd already removed it, and just pushed an update that actually does so, thanks for the quick look), but this patch doesn't touch ipv6.

hannesm:
cool, another remark: why does lots of things live under the `Make` functor in ICMP (I know it was the same in the IPv4 module)?  I'd appreciate to move code which does not need anything from `IP` to be in the main module, and have only the things which depend on the `IP` inside of the functor... (raising this here since it is fresh code, imho the old code should be refactored)...

yomimono:
Do you have a strong opinion on whether such code should be inside the Icmpv4 module but outside the Make functor, or somewhere else?  Your point is well taken and I'm happy to move it (and if you have no opinion I'll leave it in Icmpv4).

hannesm:
I think the icmp module is well-suited... or maybe wire_structs?  it might allow us to reuse the code in other libraries (such as sniffers, pcap replay, ...)

avsm:
Agreed -- minimising the amount of stuff under `Make` would be good (basically, everything that doesn't use `IP.t` can be lifted outside).  `Wire_structs` is ok for the cstruct, but somewhere else for the parsing/printing logic.  

It would be really cool if we could build a tlsvis-style tool without the `Make` functor using just the functional printers/parsers from the various protocol stack libraries.

hannesm:
Wire_structs already has a `parse_ethernet_frame`... but I agree this should live in (a fresh module) `Wire_(un)parser`.. the printing/formatting logic then in `Wire_formatter`!?  I tend to appreciate multiple small modules rather than big ones.. (but the question remains whether there should be a library+module per protocol (ethernet/IP/TCP/ARP), or per task (read/write/print))?

yomimono:
Per-protocol makes much more sense to me, FWIW.

nojb:
Note that ICMP6 cannot be separated so easily in the case of IP6 because it is used to implement address resolution (i.e. it plays the role of ARP in the v6 case)...

yomimono:
@nojb: Fair point.  We could probably still split parse/print functions for ICMPV6 out into a separate module even if the functorization isn't as clear for the bit we sandwich into the stack, I think.

@hannesm, @avsm: any more concrete thoughts on where to put this stuff?  ARP has its cstructs in a separate Arpv4_wire ; we could rearrange things with Icmpv4_wire, Icmpv4_parse, Icmpv4_print, and the Make functor for use in the stack as a template for other stuff to use?

hannesm:
@yomimono what about a directory `icmp` with the mentioned modules (`Icmpv4_parse` / `_wire` / `_print` (or format instead of print?))?

yomimono:
Having heard no objections (and having none myself) to a subdirectory per protocol, I've updated this PR with such an arrangement.  I'll submit a separate PR with this change made to the other protocol implementations.

hannesm:
this looks good to me! thanks!

lib/icmp/icmpv4.ml
```diff
@@ -0,0 +1,87 @@
...
+    | 7  -> "Destination host unknown"
+    | 8  -> "Source host isolated"
+    | 9  -> "Network administratively prohibited"
+    | 10 -> "Host administratively prohibited"
+    | 11 -> "Network unreachable for TOS"
+    | 12 -> "Host unreachable for TOS"
+    | 13 -> "Communication administratively prohibited"
+    | 14 -> "Host Precedence Violation"
+    | 15 -> "Precedence cutoff in effect"
+    | code -> Printf.sprintf "Unknown ICMP code: %d" code
```
hannesm: this can happily life outside of the functor...

yomimono: It's not exposed in the public interface for the functor, and I'd like to eventually remove it anyway.  I'd like to defer this change to another PR.

hannesm: sounds good to me as well

talex5: Seems like a useful thing to have. What are you planning to replace it with?

yomimono: I hear someone did a bunch of work on logging recently, maybe some of that is reusable ;)

hannesm: (imho an indirection is missing, a cenum from number to icmp type, and the formatter should use the high-level icmp type, rather than numbers... but this can all wait for another PR)

hannesm:
(this also deserves an entry in changes.md)

PR #199, opened 2016-05-03
Use Logs library for logging

talex5:
Note: to avoid breaking the API, this patch does not yet remove the (now-unused) console arguments from the signatures. This is so that this patch can be merged quickly to avoid creating more conflicts with other refactoring work going on.


yomimono:
Looks great to me! Thanks!

hannesm:
this is great! I'm uncertain about choice of log levels (should we come up with some guidelines? is info a reasonable default for all the messages?). and there's still a custom logging framework in tcp https://github.com/mirage/mirage-tcpip/blob/master/tcp/log.ml -- which should now use logs, shouldn't it?

talex5:
I kept most of them at info to avoid changing the behaviour too much in this PR, but feel free to make additional PRs to discuss adjustments to them.

The pre-existing logging does some stats too, so might need a bit of adjusting to move it over (but would be welcome I think). @samoht knows more about this...

PR #202, opened 2016-05-04
Log exceptions accepting TCP connections

talex5:
Before, these were passed back to the network driver. However, tcpip is
in a better position to report useful information. Example:

```
[ERROR] uncaught exception accepting new TCP flow on port 80:
        (Failure "Test exception")
```

See: https://github.com/mirage/mirage-net-xen/issues/39


tcp/pcb.ml
```diff
@@ -479,7 +486,13 @@ struct
         (* Finish processing ACK, so pcb.state is correct *)
         Rx.input t pkt newconn >>= fun () ->
         (* send new connection up to listener *)
-        pushf (fst newconn)
+        let flow = fst newconn in
+        Lwt.catch (fun () -> pushf flow)
```
talex5: Should we add a `Lwt.async` here too? I'm not sure, but it looks like we run the entire TCP flow handler in the context of the original network frame. We should probably return sooner, to let netif know we're done with the ack frame.

tcp/pcb.ml
```diff
@@ -465,6 +468,10 @@ struct
       Lwt.return_unit
     | None ->
       Tx.send_rst t id ~sequence ~ack_number ~syn ~fin
+    | exception ex ->
+      PCB_Log.err (fun f -> f "uncaught exception looking up handler for TCP port %d:@ %s"
+                      id.WIRE.local_port (Printexc.to_string ex));
+      Lwt.return_unit
```
hannesm: afaics this case can never happen -- the passed in `listeners` is from [direct stack](https://github.com/mirage/mirage-tcpip/blob/a54f6385a433407a7b29391b148967af66d4f22a/lib/tcpip_stack_direct.ml#L142-L144) and a `try` around `Hashtbl.find`.

yomimono: This is quibbling, but this `~listeners` isn't necessarily from the direct stack.  It's an internal hashtable carried in the record t ( https://github.com/mirage/mirage-tcpip/blob/master/tcp/pcb.ml#L84 ).  I think the relevant guards are here, in `new_server_connection`: https://github.com/mirage/mirage-tcpip/blob/master/tcp/pcb.ml#L369

talex5: Isn't this a public entrypoint defined in mirage-types?

https://github.com/mirage/mirage/blob/76575138f27ccfbff2a073ba997c54d4afb84149/types/V1.mli#L661

(using a stack is optional, but in that case it shouldn't fail indeed)

hannesm: well, we can decide to be completely defensive in all external APIs, but imho this would make the code a mess (and likely does checking more than once).  I'd rather strive for 'no exceptions at module boundaries', and thus if a client provides some function which fails, we fail hard.  if there are potential errors, we should use a `result` type. in this specific case I think an `option` is a good solution.

what should be handled defensively is input from the network.

talex5: Well, we did fail hard and two people decided to stick with an older version rather than fix their code :-( And there's nowhere to return an error to here because the callback is called asynchronously when a network frame arrives. If you throw here, the network layer will dump out the frame that caused the problem, but we know the problem is the port number at this point. Anyway, I don't mind getting rid of this. It was just one of the two places the user's exception might have reached, but it was probably the other one.

hannesm: @talex5 as I mentioned in the mirage-net-xen issue, I'm in favour of failing hard (my goal is to not have any exceptions in the interface).  but let us do that a bit more organised: have the same semantics in mirage-net-unix and mirage-net-xen, and fix our own (well, MirageOS) libraries in the same release cycle.

PR #203, opened 2016-05-05
spring cleanup: no need to depend on mirage-net-unix

hannesm:
these functor applications are no longer needed
(tested with mirage-skeleton, unix and tap device)


hannesm:
I couldn't find anything depending on those findlib libraries, but maybe I'm wrong?  mirage operations are fine... @samoht @avsm @yomimono any idea/reason to keep them around?

yomimono:
I think there's no longer any way to get at these with an invocation of the `mirage` front-end tool.  `unix`, `xen`, and `socket` used to be distinct modes which would yield different stacks, but we now have only the `direct` and `socket` stacks.

hannesm:
travis seems to agree... fragments from a `main.ml` generated for unix direct stack:

```
module Ethif1 = Ethif.Make(Netif)
module Arpv41 = Arpv4.Make(Ethif1)(Clock)(OS.Time)
module Ipv41 = Ipv4.Make(Ethif1)(Arpv41)
module Icmpv41 = Icmpv4.Make(Ipv41)
module Udp1 = Udp.Make(Ipv41)
module Tcp1 = Tcp.Flow.Make(Ipv41)(OS.Time)(Clock)(Random)
module Tcpip_stack_direct1 = Tcpip_stack_direct.Make(Console_unix)(OS.Time)
  (Random)(Netif)(Ethif1)(Arpv41)(Ipv41)(Icmpv41)(Udp1)(Tcp1)
```

thus I think it is safe to remove (likely even the `mirage-unix` depedency is not needed in the opam file -- will add another commit to do so?)

talex5:
Nice find! It does look like Mirage does all this itself now.

hannesm:
travis is happy... so am I

hannesm:
(tried to let travis build with 4.03, but since pcap-format doesn't yet work with 4.03, this failed.. but 4.03 build can very well be another PR at some point)

PR #204, opened 2016-05-14
Udp: set the checksum field to 0 before calculating the checksum

djs55:
The Io_page library will call `memset` and returned zero memory, but set the checksum to 0 explicitly to avoid future misunderstandings.

### DIFF
--- a/lib/udp.ml
+++ b/lib/udp.ml
@@ -60,7 +60,7 @@ module Make(Ip: V1_LWT.IP) = struct
     Wire_structs.set_udp_source_port udp_buf source_port;
     Wire_structs.set_udp_dest_port udp_buf dest_port;
     Wire_structs.set_udp_length udp_buf (Wire_structs.sizeof_udp + Cstruct.lenv bufs);
-    (* Wire_structs.set_udp_checksum udp_buf 0; *)
+    Wire_structs.set_udp_checksum udp_buf 0;
     let csum = Ip.checksum frame (udp_buf :: bufs) in
     Wire_structs.set_udp_checksum udp_buf csum;
     Ip.writev t.ip frame bufs


talex5:
io-page should zero it here:

https://github.com/mirage/io-page/blob/77dd8cf1ef0472df365361d74f09d14e6999ecc1/lib/stub_alloc_pages.c#L72

Have you seen it fail to do this?

talex5:
(I think any function that returns uninitialised memory should have an `_unsafe` suffix or similar, since it's a likely source of private data being leaked)

djs55:
Ah I've actually been running with a patch which turns `Io_page.get` into a synonym for `Cstruct.create` on Windows which is why I had a problem.

I like the idea of an `_unsafe` suffix or similar.

What shall we do about this particular line of code? I don't think having it commented out is sensible -- it should either be deleted altogether or uncommented IMHO. If it was deleted then we run the risk of having this conversation again :) so a comment saying why the line is missing would be justified. Then again, just reinstating the line is obviously correct and makes the code more robust to being moved around (e.g. if someone copied it but fed it a raw `Cstruct.create` rather than an `Ip.allocate_frame` it would cause occasional corruption.) WDYT?

talex5:
I'm happy with either applying this patch or adding a comment. I just wanted to check io-page was working correctly. Writing an extra 16 bits isn't going to slow us down much.

Long-term, I propose:

- Change `Cstruct.create` to zero things out, and add `Cstruct.create_unsafe` for the current behaviour.

- Have the cstruct syntax extension generate a function which takes all the fields as arguments and thus ensures they're all filled in, rather than requiring the user to remember to call each setter. If it also allocated, then it could use the faster `create_unsafe` version, knowing it was about to overwrite everything.

avsm:
> Have the cstruct syntax extension generate a function which takes all the fields as arguments and thus ensures they're all filled in, rather than requiring the user to remember to call each setter. If it also allocated, then it could use the faster create_unsafe version, knowing it was about to overwrite everything.

This sounds like an excellent plan -- looks like mirage/ocaml-cstruct#30 already exists.

djs55:
OK, I'll update the misleading commit comment, and commit the resulting patch.

Issue #201, opened 2016-05-04
Unit-tests pass if an exception is thrown


talex5:
In `vnetif_common.ml`:

        | Unix.Unix_error _ ->
          Printf.printf "Could not create pcap file %s - something along the way doesn't exist.\n" pcap_file;
          Lwt.return_unit

The `return_unit` means that the test will report as passed in this case. Using `fail` instead, I see:

```
Could not create pcap file tests/pcap/tcp_connect_two_stacks_basic.pcap - something along the way doesn't exist.
[exception] Unix.Unix_error(Unix.ENOENT, "open", "tests/pcap/tcp_connect_two_stacks_basic.pcap")
```

I noticed this because I was trying to get the tests to fail, but they passed no matter what...

talex5:
Also, from #205 (so it doesn't get lost):
```
+    with
+    | Invalid_argument s -> Printf.printf "Arpv4_wire failed on buffer: %s" s;
+      Lwt.return_unit
```

PR #205, opened 2016-05-17
Disaggregate protocols, and expose parsing and printing.

yomimono:
This PR requires a small change to mirage-types -- module type IP must provide a `pseudoheader` function.  See https://github.com/mirage/mirage/pull/541 for details.

This PR moves all protocol implementations into their own directories in `lib`.  It also makes several structural changes:

* protocols now have their own `Protocol_wire` modules.  The top level `Wire_structs` module no longer exists.
* protocols expose and use `Protocol_parse` and `Protocol_print` modules.  These provide convenience functions for extracting meaning from `Cstruct.t`s and constructing `Cstruct.t`s given the relevant information.  This division is intended to make it easier for users to construct modules with different logic from the stock implementations without needing to reimplement or copy serialization and deserialization logic.
* The `logs` library is used instead of the custom TCP `Log` module.
* Use of `Result` has replaced polymorphic result types and exceptions in many locations.

This PR is large, but hopefully provides many benefits over the current state of affairs.  I'm thankful for any attention you can give to it.


yomimono:
The lack of attention given to `lib/dhcp` in this PR is intentional; it's likely time to replace the implementation there with calls to `charrua-core`.  I'd like to do that in a separate PR as this one is already, um, large.

hannesm:
I think a dhcp client might also very well deserve its own repository, and does not need to be bundled with the tcp/ip stack

.travis.yml
```diff
@@ -3,7 +3,7 @@ script: bash -ex .travis-ci.sh
 sudo: required
 env:
   global:
-  - PINS="mirage:https://github.com/mirage/mirage.git mirage-types:https://github.com/mirage/mirage.git"
+          - PINS="mirage:https://github.com/mirage/mirage.git mirage-types:https://github.com/yomimono/mirage.git#pseudoheaders mirage-types-lwt:https://github.com/yomimono/mirage.git#pseudoheaders"
```
hannesm: there's some additional spaces in the front? intentional?

yomimono: What a helpful editor I have!

lib/ethif/ethif.ml
```diff
@@ -38,18 +39,19 @@ module Make(Netif : V1_LWT.NETWORK) = struct
...
-        | Some Wire_structs.IPv4 -> ipv4 payload
-        | Some Wire_structs.IPv6 -> ipv6 payload
-        | None -> Lwt.return_unit (* TODO: default ethertype payload handler *)
+        let open Ethif_wire in
+        match header.ethertype with
+        | ARP -> arpv4 header.payload
+        | IPv4 -> ipv4 header.payload
+        | IPv6 -> ipv6 header.payload
       end
     | _ -> Lwt.return_unit
```
hannesm: here I'd expect a log message as well?

hannesm:
in your parsers you define the `type t` (containing mostly high-level types), I like this (and would like to see even more high-level things (e.g. ``proto : [`ICMP|`TCP|`UDP]``).. shouldn't the printer use this very same `t`?
another question is whether the printer should allocate its buffer (which would allow to return just a `Cstruct.t` instead of a `unit result` -- but it's fine as it is now (and likely a more intrusive change depending on the allocation pattern of the library)

Makefile
```diff
@@ -39,16 +39,3 @@ configure:
 .PHONY: build doc test all install uninstall reinstall clean distclean configure
 
 # OASIS_STOP
```
talex5: Where did the release target go? Is it no longer used? I didn't see it mentioned in the changes.

lib/arpv4/arpv4.ml
```diff
@@ -94,68 +87,53 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
...
-    let dmac = Macaddr.to_bytes arp.tha in
-    let smac = Macaddr.to_bytes arp.sha in
-    let spa = Ipaddr.V4.to_int32 arp.spa in
-    let tpa = Ipaddr.V4.to_int32 arp.tpa in
-    let op =
+    match Arpv4_print.print_arpv4_header
+      ~buf:(Cstruct.shift buf Ethif_wire.sizeof_ethernet)
+      ~src_ip:arp.spa ~dst_ip:arp.tpa ~src_mac:arp.sha ~dst_mac:arp.tha
+      ~op:arp.op with
+    | Error s -> Log.info (fun f -> f "Failed to print Arpv4 header: %s" s);
```
talex5: Shouldn't this be more serious than an info-level log message? Since we allocate the buffer just above, presumably at the right size, we should probably `failwith s` here to indicate a bug.

lib/arpv4/arpv4.ml
```diff
@@ -94,68 +87,53 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
...
+        Ethif.write t.ethif buf
+
+  (* Input handler for an ARP packet *)
+  let input t frame =
+    let open Arpv4_parse in
+    MProf.Trace.label "arpv4.input";
+    match parse_arpv4_header frame with
+    | Result.Error s ->
+      Log.info (fun f -> f "Failed to parse arpv4 header: %a (buffer: %S)"
+                   Arpv4_parse.pp_error s (Cstruct.to_string frame));
```
talex5: (possibly useful: Cstruct 2.1.0 has a `hexdump_pp` function)

lib/arpv4/arpv4.ml
```diff
@@ -175,7 +153,7 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
     (* Source protocol address, pick one of our IP addresses *)
     let spa = match t.bound_ips with
       | hd::_ -> hd | [] -> Ipaddr.V4.any in
-    output t { op=`Request; tha; sha; tpa; spa }
+    output t Arpv4_parse.({ op=Arpv4_wire.Request; tha; sha; tpa; spa })
```
talex5: It looks a bit awkward using the parser module's type to generate output. Perhaps parsing and printing should be in a single `Arpv4_msg` module? In an ideal world, we might even generate the parser and the printer from a single definition.

hannesm:
after sleeping, I still prefer `Printer.print : t -> Cstruct.t` over `print : buf -> x -> y -> z -> unit result` less complexity (and control flow), more local understanding -- and in the end you need to append those headers anyways...

lib/dhcp/dhcp_clientv4.ml
```diff
@@ -79,23 +79,18 @@ module Make
     let options = Dhcpv4_option.Packet.to_bytes options in
     let options_len = Bytes.length options in
     let total_len = options_len + sizeof_dhcp in
-    let buf = Io_page.(to_cstruct (get 1)) in
+    let buf = Cstruct.create total_len in
+    Cstruct.memset buf 0x00;
```
talex5: Could add a comment that this won't be needed with the next cstruct release.

lib/ethif/ethif.ml
talex5: Yes, would be good, although this bit isn't part of this PR.

lib/icmp/icmpv4.ml
```diff
@@ -43,30 +22,37 @@ module Make(IP : V1_LWT.IPV4) = struct
...
       Lwt.return_unit
+    | Result.Ok message ->
+      let open Icmpv4_wire in
+      let open Icmpv4_parse in
+      match message.ty, message.subheader with
+      | Echo_reply, _ -> Log.info (fun f -> f "ICMP: discarding echo reply from %a" Ipaddr.V4.pp_hum src);
+        Lwt.return_unit
+      | Destination_unreachable, _ ->
+        Log.info (fun f -> f "ICMP: destination unreachable from %a" Ipaddr.V4.pp_hum src);
+        Lwt.return_unit
```
talex5: I guess we should add a `TODO` here. Presumably ICMP results should go back to the application somehow?

lib/icmp/icmpv4_parse.ml
```diff
@@ -17,25 +18,34 @@ type t = {
...
+    | Some ty -> Result.Ok ty
+  in
+  check_len () >>= check_ty >>= fun ty ->
+  let code = get_icmpv4_code buf in
+  let csum = get_icmpv4_csum buf in
+  let subheader = subheader_of_cstruct ty (Cstruct.shift buf 4) in
+  let payload =
+    if Cstruct.len buf > sizeof_icmpv4
+    then Some (Cstruct.shift buf sizeof_icmpv4)
+    else None
```
talex5: (not changed in this PR but...) Does this need to be an option type? If we shift in all cases we'll just get an empty cstruct, which is fine right?

lib/ipv4/ipv4.ml
```diff
@@ -97,61 +97,66 @@ module Make(Ethif: V1_LWT.ETHIF) (Arpv4 : V1_LWT.ARP) = struct
...
+    set_ipv4_len buf tlen;
+    set_ipv4_id buf (Random.int 65535); (* TODO *)
+    set_ipv4_csum buf 0;
     let checksum = Tcpip_checksum.ones_complement buf in
-    Wire_structs.Ipv4_wire.set_ipv4_csum buf checksum
+    set_ipv4_csum buf checksum
 
   let allocate_frame t ~dst ~proto =
+    let open Ipv4_wire in
     let ethernet_frame = Io_page.to_cstruct (Io_page.get 1) in
```
talex5: (Do we still want to use io-pages everywhere? I saw an io-page converted to `Cstruct.create` above in the dhcp code.)

lib/ipv4/ipv4.ml
```diff
@@ -97,61 +97,66 @@ module Make(Ethif: V1_LWT.ETHIF) (Arpv4 : V1_LWT.ARP) = struct
...
+    | Error s -> 
+      Log.info (fun f -> f "IP.allocate_frame: could not print ethernet header: %s" s);
+      (ethernet_frame, len)
+    | Ok () ->
+      let buf = Cstruct.shift ethernet_frame Ethif_wire.sizeof_ethernet in
+      (* TODO: why 38 for TTL? *)
+      match Ipv4_print.print_ipv4_header ~buf ~src:t.ip ~dst ~proto ~ttl:38 with
+      | Error s ->
+        Log.info (fun f -> f "IP.allocate_frame: could not print IPv4 header: %s" s);
+        (ethernet_frame, len)
```
talex5: Shouldn't we return an error here?

lib/tcp/pcb.ml
```diff
@@ -493,47 +457,49 @@ struct
...
+        let dest_port = Tcp_wire.get_tcp_dst_port data in
+        let id = WIRE.wire ~local_port:dest_port ~dest_port:source_port ~dest_ip:src ~local_ip:dst in
+        (* Lookup connection from the active PCB hash *)
+        with_hashtbl t.channels id
+          (* PCB exists, so continue the connection state machine in tcp_input *)
+          (Rx.input t data)
+          (* No existing PCB, so check if it is a SYN for a listening function *)
+          (input_no_pcb t listeners data)
+      with
+      | Invalid_argument s ->
```
talex5: I think this `try` is wrapping a bit too much code.

lib/tcp/window.ml
```diff
@@ -108,25 +109,23 @@ let t ~rx_wnd_scale ~tx_wnd_scale ~rx_wnd ~tx_wnd ~rx_isn ~tx_mss ~tx_isn =
...
 
-(* Check if a sequence number is in the right range
-   TODO: modulo 32 for wrap
-*)
+(* Check if a sequence number is in the right range *)
 let valid t seq =
   let redge = Sequence.(add t.rx_nxt (of_int32 t.rx_wnd)) in
   let ledge = Sequence.(sub t.rx_nxt (of_int32 t.max_rx_wnd)) in
   let r = Sequence.between seq ledge redge in
   (* PERF: ~5% perf degradation if commenting out that line
```
talex5: Out-of-date comment?

lib/udp/udp.ml
```diff
@@ -36,39 +42,44 @@ module Make(Ip: V1_LWT.IP) = struct
 
   let id {ip} = ip
```
talex5: I don't think this is used any longer. It's probably worth adding `true: warn(A-4)` to the `_tags` file to catch unused code (and other problems).

lib/udp/udp.ml
```diff
@@ -36,39 +42,44 @@ module Make(Ip: V1_LWT.IP) = struct
...
+    | Error s -> Log.debug (fun f -> f "Discarding transmitted UDP message: error writing: %s" s);
+      Lwt.return_unit
 
   let write ?source_port ~dest_ip ~dest_port t buf =
     writev ?source_port ~dest_ip ~dest_port t [buf]
 
-  let connect ip = Lwt.return (`Ok { ip })
+  let connect ip =
+    let ips = List.map Ip.to_uipaddr @@ Ip.get_ip ip in
+    Log.info (fun f -> f "UDP interface disconnected on %a" pp_ips ips);
```
talex5: `disconnected`?

lib_test/test_icmpv4.ml
```diff
@@ -57,9 +57,13 @@ end = struct
 
   let input t buffer =
     (* disregard responses, but reply to queries *)
+    try
     match Arpv4_wire.get_arp_op buffer with
     | 1 -> A.input t.base buffer
     | 2 | _ -> Lwt.return_unit
+    with
+    | Invalid_argument s -> Printf.printf "Arpv4_wire failed on buffer: %s" s;
+      Lwt.return_unit
```
talex5: Why are we ignoring this error?

lib_test/test_icmpv4.ml
```diff
@@ -78,6 +82,15 @@ type stack = {
...
 }
 
+let testbind x y =
+  match x with
+  | Result.Ok p -> y p
+  | Result.Error s -> Alcotest.fail s
+let (>>=?) = testbind
+
+let slowly fn =
+  OS.Time.sleep 0.1 >>= fun () -> fn >>= fun () -> OS.Time.sleep 0.1
```
talex5: Note that `fn` starts before `slowly` is called. This function waits 0.1s, then waits for `fn` to complete, then waits another 0.1s. Probably it should be `fn ()`. Also, possibly it should `Lwt_unix.yield` instead of sleeping? What is it for?

setup.ml
```diff
@@ -1,9 +1,9 @@
 (* setup.ml generated for the first time by OASIS v0.4.5 *)
 
 (* OASIS_START *)
-(* DO NOT EDIT (digest: b8a8e2c61af2e8d266d45d6fbc45851d) *)
+(* DO NOT EDIT (digest: af08e5f6f834fc806bb229d1fdc89820) *)
 (*
-   Regenerated by OASIS v0.4.6
+   Regenerated by OASIS v0.4.5
```
talex5: Need to use 0.4.6 to support OCaml 4.03?

talex5:
This looks great! Due to all the code motion, it was difficult to see what was the same as before so I just skimmed and commented on whatever I noticed. This means that most of my comments are probably about pre-existing issues.

Makefile
yomimono: Thanks for noticing this - I accidentally clobbered it when committing the results of `oasis setup` for the nine zillionth time.  I've restored it.

yomimono:
Your points are well-taken about `t` in the printers.  I thought it would be a bit cumbersome to write the printers this way, since generally the logic will be that the caller constructs and assembles the necessary components of `t` and doesn't particularly want to do much with them as a unit besides emit them on the wire.  In that case, you end up having to call `make` or directly construct a record just to send it to the printing function, which feels a bit cumbersome to me.

_oasis
```diff
@@ -35,84 +35,78 @@ Library tcpip_xen
...
 Library arpv4
   CompiledObject:     best
-  Path:               lib
+  Path:               lib/arpv4
   Findlibparent:      tcpip
   Findlibname:        arpv4
-  Modules:            Arpv4, Arpv4_wire
-  BuildDepends:       tcpip,io-page,mirage-types,ipaddr,cstruct,cstruct.ppx,lwt
+  Modules:            Arpv4, Arpv4_wire, Arpv4_parse, Arpv4_print
+  BuildDepends:       mirage-types,logs,ipaddr,cstruct,cstruct.ppx,lwt,
```
samoht: this would probably requires a XMETArequires line to not depend on `ppx` at runtime. Note: I'm not totally sure, so worth checking with @djs55 who had this issue recently.

_oasis
```diff
@@ -35,84 +35,78 @@ Library tcpip_xen
...
 Library ipv4
   CompiledObject:     best
-  Path:               lib
+  Path:               lib/ipv4
   Findlibparent:      tcpip
   Findlibname:        ipv4
-  Modules:            Ipv4
-  BuildDepends:       io-page,mirage-types,ipaddr,cstruct,lwt,tcpip
+  Modules:            Ipv4, Ipv4_wire, Ipv4_parse, Ipv4_print
+  BuildDepends:       logs,io-page,mirage-types,ipaddr,cstruct.ppx,lwt,result,rresult,
```
samoht: same here (`cstruct.ppx`)

samoht:
Splitting the library is a great, thanks for the work!

Can you also add the missing `mli`? I'd make code review much more easy (at least for me :p) ... and ideally with some comments in it :p That would be very helpful to see how the APIs fit together.

Also, I'm not sure to like very much the `_print` and `_parse` conventions. Maybe `_write` and `_read`?

lib/ipv4/ipv4.ml
yomimono: Here, we're trying to allocate "enough" memory for whatever the caller's data might be, without knowing how much that is, so a page seems like a reasonable amount.  Do you think changing it to Cstruct.create 4096 is materially different?

(I'd like to remove this function entirely, and change the "ip allocates a buffer, other modules write into it at the provided offset" to something else, but that's a rather large API change)

lib/ipv4/ipv4.ml
yomimono: I'd like to, but `V1.mli

yomimono: Yes, but `V1.mli` provides no facility for `allocate` to return an error; we'd have to throw an exception.

yomimono:
I'm totally fine to change `_print` and `_parse` to something else, but we should probably decide once and leave it that way.  @talex5, I seem to recall that you proposed `marshal` and `unmarshal` -- what's your opinion?

lib/icmp/icmpv4.ml
yomimono: I think in the general case, they probably shouldn't -- reliable transport protocols have their own ways of dealing with these failures, and unreliable ones don't care (or have something in-band at the application level to detect and handle failures).

I think if you care about ICMP results, you can write and use an ICMP module that throws an exception or responds to messages like this, but informing the user out-of-band is a reasonable default.  Thoughts, @hannesm @avsm ?

lib/icmp/icmpv4_parse.ml
yomimono: It's purely a preference for having nothing be indicated by `None` rather than something empty.  It sounds like you'd prefer the Cstruct of size 0?

lib/tcp/pcb.ml
yomimono: On looking at this again it's not clear to me why this is in its own `try` at all; I've added the ports to the record returned by the parse function and now `input` just runs `parse` right away.  Thanks for noticing this.

talex5:
`read` and `write` works for me. This is also the convention ocaml-9p uses, with one module per message type, e.g.

https://github.com/mirage/ocaml-9p/blob/cb9f5a6c83051ac5dfa7fc7a43ca6a2910a0a241/lib/protocol_9p_request.ml#L23

"print" and "parse" both vaguely suggest text formats to me.

lib/icmp/icmpv4_parse.ml
talex5: Yes. Otherwise (from the type system's pov) both None and the empty cstruct are possible values, but one of them doesn't mean anything. Might as well make that confusion impossible, unless there's some strong reason why application code should handle None in a different way to an empty cstruct?

hannesm:
`read` and `write` always remind me of effectful operations, and I'd avoid them here (and use instead `parse` and `assemble` or `marshal` and `unmarshal`).

lib/ipv4/ipv4.ml
talex5: OK, makes sense now (and I agree with changing the API in the long term).

lib/ipv4/ipv4.ml
talex5: That's probably better than returning an invalid buffer (and I guess there's a bug if we get here anyway).

lib/arpv4/arpv4.ml
```diff
@@ -63,8 +55,9 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
         | Pending _ -> expired
         | Confirmed (t, _) -> if t >= now then ip :: expired else expired) t.cache []
     in
-    List.iter (fun ip -> Log.info (fun f -> f "ARP: timeout %a" Ipaddr.V4.pp_hum ip)) expired;
-    List.iter (Hashtbl.remove t.cache) expired;
+    List.iter (fun ip ->
+         Log.info (fun f -> f "ARP: timeout %a" Ipaddr.V4.pp_hum ip); Hashtbl.remove t.cache ip
```
hannesm: style: newline after `;`

lib/ethif/ethif.ml
```diff
@@ -16,6 +16,10 @@
  *
  *)
 open Lwt.Infix
+open Result
+
+let src = Logs.Src.create "ethif" ~doc:"Mirage Ethernet"
```
hannesm: I'm not sure about this... shouldn't we create the log instance in `connect` (and include some name)?   this way, if we have multiple ethernet interfaces, we can distinguish the individual log messages

lib/icmp/icmpv4.ml
hannesm: in UNIX sockets, your `read/write` get errors received via ICMP frames, which afaict is not done in MirageOS at all... not sure how to feed an ICMP error message up the TCP/UDP stack... but clearly we should have such a signalling path

lib/icmp/icmpv4_print.ml
```diff
@@ -1,10 +1,58 @@
...
+let echo_reply ~buf ?payload ~id ~seq =
+  echo ~buf ?payload ~ty:Echo_reply ~id ~seq
+
+(** [would_fragment ip_header ip_payload next_hop_mtu] generates an
+    ICMP destination unreachable message, with the code set to 4 ("packet
+    fragmentation is required but the don't-fragment bit is set").  [ip_header] should
+    be the IP header of the packet which will be rejected; [ip_payload] is the
+    optional first 8 bytes of the IPv4 payload (generally the first 8 bytes of
+    a UDP or TCP header).  Payloads of length > 8 will be truncated. *)
+let would_fragment ~buf ~ip_header ?ip_payload ~next_hop_mtu =
```
hannesm: this is new functionality, not used anyware afaics?

lib/ipv4/ipv4_parse.ml
```diff
@@ -0,0 +1,48 @@
...
+  let parse buf options_len =
+    try
+      let src = Ipaddr.V4.of_int32 (get_ipv4_src buf) in
+      let dst = Ipaddr.V4.of_int32 (get_ipv4_dst buf) in
+      let proto = get_ipv4_proto buf in
+      let options =
+        if options_len > 0 then Some (Cstruct.sub buf sizeof_ipv4 options_len)
+        else None
+      in
+      let payload_len = (get_ipv4_len buf) - sizeof_ipv4 - options_len in
```
hannesm: isn't there some check needed that the buffer length is payload_len + header? otherwise out of bounds accesses...

mor1:
Re `print`/`parse` vs `read`/`write` -- I probably prefer marshal/unmarshal (over all of print/parse, read/write and serialise/deserialise). print, read, write all sound like invoking actual IO operations to me rather than simply data manipulations. I've also used things like `of_buf` / `to_buf` and similar in the past I think.

lib/icmp/icmpv4_print.ml
yomimono: See https://github.com/docker/vpnkit/blob/master/src/hostnet/lib/slirp.ml#L74 for an example.  A proper NAT implementation would have to include this (among other ICMP pieces) as well.

hannesm: ah, great. doesn't hurt to add here :)

hannesm:
it looks like `marshal/unmarshal` is a reasonable consensus, let's use those (it'll take some time till I manage to rename the parsers).. we might even want to define a reasonable module type for a generic network dumper etc.  (but marshal/unmarshal shouldn't stop this from getting merged, can well be another PR)

dbuenzli:
> it looks like marshal/unmarshal is a reasonable consensus,

This reminds me to much of the `Marshal` module. I would rather use `encode` and `decode`.

mor1:
>> it looks like marshal/unmarshal is a reasonable consensus,
> This reminds me to much of the `Marshal` module. I would rather use encode and decode.

Hadn't come across the `Marshal` module before-- how widely used is it?
`marshal`/`unmarshal` make more sense to me in a networking context than `encode`/`decode`.

yomimono:
`marshal/unmarshal` is also used in ocaml-dns, and I agree with @mor1 about it vs encode/decode.  A final update to this PR with those changes is coming today; it can either stand or fall after that.

yomimono:
Most comments are addressed in the changeset just pushed.  If there are additional changes needed before folks think this is good to merge, I'll close this PR and open a new one.

lib/ethif/ethif.ml
```diff
@@ -42,12 +42,11 @@ module Make(Netif : V1_LWT.NETWORK) = struct
   let mac t = Netif.mac t.netif
 
   let input ~arpv4 ~ipv4 ~ipv6 t frame =
-    let open Ethif_parse in
     MProf.Trace.label "ethif.input";
     let of_interest dest =
       Macaddr.compare dest (mac t) = 0 || not (Macaddr.is_unicast dest)
     in
-    match parse_ethernet_header frame with
+    match Ethif_unmarshal.of_cstruct frame with
```
talex5: Having marshal and unmarshal in separate modules still reads badly to me. `Ethif.of_cstruct` would make more sense.

talex5:
Looks pretty good - thanks!

What about the comments on `lib_test/test_icmpv4.ml` though? If a unit-test fails, we should abort, not just print to the console (Travis won't notice that).

Also (might just be personal preference) but I'd prefer to have e.g. `Udp_packet.of_cstruct` rather than `Udp_unmarshal.of_cstruct`, which sounds like we're creating an unmarshaller rather than a UDP packet. It seems more common in OCaml to have one module per type, with marshal and unmarshal as functions inside it.

lib/ethif/ethif.ml
mor1: Hm-- to some extent I'm inclined to go with this one and then think further about naming of things. Splitting out the modules seems a good thing regardless of more name wrangling required. And possibly having things split into smaller modules will make it easier to think coherently about what the API should be...

(FWIW I had though that `marshal`/`unmarshal` were to be function names inside something like `Tcp4_wire` rather than embedded directly in module names -- apologies for any noise contributed thereby.)

talex5: Sure; it was just a suggestion

yomimono:
It seems that the comments here are that this change is generally OK, although the split of modules between `_unmarshal` and `_marshal` is weird, and that there are a few issues with the tests that should be fixed (see also #201).  I've pushed a change to make the tests pass with Logs 0.6.0, but I'd like to address naming concerns and test improvements in a separate PR.  Can I get a thumbs-up/thumbs-down on the current state of the PR?  I'd like to merge or close by the end of the week.

mor1:
Because I don't trust emoji-based communication alone (I'm too old...) -- yes, that seems good to me!

PR #209, opened 2016-06-09
_{un}marshal -> _packet; payloads removed from _packet.t

yomimono:
Implementations of some suggestions based on #205 .  Take the `_marshal` and `_unmarshal` modules for each protocol and make them submodules of a `_packet` module; marshalling and unmarshalling deals with a type `t` defined at the `_packet` level.

Additionally, *finally* remove the custom ARP parser from the testing code -- one of the first duplicated custom parsers I hope to `git rm`.


samoht:
LGTM

mor1:
Looks good -- definite improvement! 

Three queries: 

1. I wonder about the `make_cstruct` vs `to_cstruct` distinction though -- at least in the Arp case it seems that there's duplicated code between the two. 
2. I think it would probably also seem more logical to me to refactor the API to something like `val t_of: ...parameters... -> t`, and `val to_cstruct: t -> Cstruct.t`, though I'm also wondering about splitting the latter so that there's one function that takes the buffer explicitly, and another that allocates a suitable buffer and calls the first. 
3. I'm also wondering about the relationship between `_packet` and `_wire` and which the minimal set of things that a consumer of this library might want if they were manipulating packets. But I don't know what the answer is to that.

hannesm:
I like this!
but I share a similar concern as @mor1 and would prefer a marshal interface:
```OCaml
val to_cstruct : t -> Cstruct.t
val into_cstruct : t -> Cstruct.t -> (unit, string) result
```
(since the `type t` is exposed and not private, there's no need for `t_of` @mor1 mentioned)

I don't see a reason to deal with payload in the marshal... either (pre-allocated) the user calls into_cstruct with a shifted cstruct or the user calls `to_cstruct` for all layers and concatenates those manually.

yomimono:
We need access to the payload in `marshal` for some protocols because we need it to correctly compute the checksum in the header.  ICMP, TCP, and UDP share this property.

hannesm:
thx for clarifying, please ignore that part of my comment ;)

yomimono:
@mor1: I've used `_wire` in some test code to purposely make slightly-wrong packets, so it at least has that use.  I'd prefer to leave it available.

@mor1: good point on the duplicated code; I'll fix that.

@hannesm and @mor1: I think `into_cstruct: t -> Cstruct.t -> (unit, error) result` is nicer too.  I'll change that.

@talex5 suggested that rather than `~payload` as an argument to the marshallers that need it, the name be something which more explicitly tells the caller that the payload won't be blitted into the returned `Cstruct.t`.  I've tried to be very clear about it in the docstring for `make_cstruct`, but it seems ripe for confusion and I'd appreciate any suggestions for names or API changes that more clearly signal that you're only getting a header back, not a header + payload.

talex5:
Yeah, it's not clear what the right API is for this. I see three possibilities:

1. `Ipv4_header.to_cstruct t ~for_payload:data` - makes it clear we're making a *header* for a particular payload, but not the payload itself. Unusual signature for a marshal function, though.

2. `Ipv4_header.to_cstruct t` where `t` contains an explicit `checksum` field. This moves the problem from the `to_cstruct` to somewhere else. It's still easy to mix a header for the wrong payload, though.

3. `Ipv4_packet.to_cstruct t` returns the header and the payload together (as `[hdr; payload]` to avoid copying). An annoyance here is we rely on the payload being immutable, but Cstruct doesn't enforce this.

talex5:
(but this PR is a nice improvement on what we have, so fine to merge as far as I'm concerned)

yomimono:
I feel strongly that (2) is not a good idea; the user shouldn't need to construct the checksum themselves.  (3) would be fine if we had immutable cstructs, but I think it's asking for trouble in a world without them.

mor1:
@hannesm: Agree re `t_of`, fair point.
@yomimono Ah-- fair enough re `_wire` then
@talex5 @yomimono I agree (2) seems a bad idea. Further variants:

1. `val ..._packet.to_cstruct: ?payload:Cstruct.t -> t -> Cstruct.t`? If `payload` is provided, it's copied after the header. An `...into_cstruct: ~buf:Cstruct.t -> ...` version could be provided where the buffer is passed in explicitly. 
2. `val ..._packet.
3. `val header: t -> Cstruct.t` and `val marshall: ?payload:Cstruct.t -> t -> Cstruct.t`, or `marshall: ?payload:Cstruct.t -> t -> Cstruct.t * Cstruct.t` where the tuple is the pair of (header, payload), for a non-copying interface. I agree it would be nice if `Cstruct.t` actually enforced read-only-ness.

I even played around with having `val t: Cstruct.t -> t` and `val h: t -> Cstruct.t` when fiddling with `pcap-format` but that might be a little too terse.

(All with appropriate error wrapping, obviously :)

talex5:
@mor1 how do your examples calculate the correct length and checksum for the header when they don't get a payload?

samoht:
(just to nitpick: `t_of: ...parameters... -> t` is not very idiomatic, usually this is called `create` or `v` -- as in value -- or the explicit name of the thing you are making)

yomimono:
Since `charrua-core` needed an update in order for the CI tests to pass, I went ahead and made a branch that uses the new marshal/unmarshal interface.  A diff is available at https://github.com/yomimono/charrua-core/commit/606bdb5ef5a6e365020e1e3ec3d4ac71911631e1 , and I very much enjoy the amount of logic it's no longer necessary to duplicate between tcpip and charrua.  Let me know what you think!

lib/icmp/icmpv4_packet.mli
```diff
@@ -25,14 +24,11 @@ end
...
   type error = string
 
-  val echo_request : buf:Cstruct.t -> payload:Cstruct.t ->
-    id:Cstruct.uint8 -> seq:Cstruct.uint8 -> (unit, error) Result.result
-
-  val echo_reply : buf:Cstruct.t -> payload:Cstruct.t ->
-    id:Cstruct.uint8 -> seq:Cstruct.uint8 -> (unit, error) Result.result
-
-  val would_fragment : buf:Cstruct.t -> ip_header:Cstruct.t ->
-    ip_payload:Cstruct.t -> next_hop_mtu:Cstruct.uint16 -> (unit, error) Result.result
```
hannesm: why did those functions disappear from the mli? if they are not needed, maybe also remove them from the ml?

lib/ipv4/ipv4_packet.ml
```diff
@@ -94,23 +97,24 @@ module Marshal = struct
     set_ipv4_proto buf t.proto;
     set_ipv4_src buf (Ipaddr.V4.to_int32 t.src);
     set_ipv4_dst buf (Ipaddr.V4.to_int32 t.dst);
-    Cstruct.blit t.options 0 buf sizeof_ipv4 (Cstruct.len t.options)
+    Cstruct.blit t.options 0 buf sizeof_ipv4 (Cstruct.len t.options);
+    set_ipv4_len buf (sizeof_ipv4 + (options_len / 4) + (Cstruct.len payload))
```
hannesm: nitpick: the `options / 4` makes me cringe.  the same expression is 6 lines above... maybe instead of `nearest_4` we should have some word-length-options?

hannesm:
I do like these changes (apart from the minor nitpicks; I guess the ICMP marshal/unmarshal and type t could get some more attention (embedding its op into t)).  nothing serious, thus pls merge if you feel like.

PR #211, opened 2016-06-13
ICMPv4 support for the socket stack

yomimono:
Provide preliminary ICMP support for the socket stack via unprivileged sockets. See https://lwn.net/Articles/420800/ or http://www.manpagez.com/man/4/icmp/ for more information.


yomimono:
The socket ICMP test fails with `EACCES` on Travis.  It may be possible to make it pass by asking for the right Linux distribution.  FWIW tests do pass locally on OSX.

unix/icmpv4_socket.ml
```diff
@@ -0,0 +1,69 @@
+open Lwt.Infix
```
talex5: `.mli` file missing from commit?

yomimono: Indeed, and including it reveals a couple of needed fixes; thanks for pointing this out.

unix/icmpv4_socket.mli
```diff
@@ -0,0 +1,5 @@
+include V1_LWT.ICMP with type ipaddr = Ipaddr.V4.t
+                     and type buffer = Cstruct.t
+                     and type 'a io = 'a Lwt.t
+
+val connect : unit -> unit io
```
talex5: `unit -> t io`?

yomimono: This PR needs more coffee.

yomimono:
Since it seems there are no objections, I'll merge this once it's passing.  ISTM that even on platforms that don't support unprivileged sockets, this is an improvement over the complete lack of support that previously existed.

PR #212, opened 2016-06-14
include a static arp table, and use it in icmpv4 tests.

yomimono:
A static ARP table has been duplicated in at least two user applications (see https://github.com/talex5/qubes-mirage-firewall/blob/1134b64f5ef0ab43c7a946619a1c77362ed81fa0/client_eth.mli#L33 and https://github.com/docker/vpnkit/blob/master/src/hostnet/lib/arp.ml ), as well as in local test code.  Provide a static ARP implementation so nobody has to duplicate it again.

Signed-off-by: Mindy Preston <mindy.preston@docker.com>


yomimono:
@talex5 and @djs55, let me know if you have any thoughts on this.

talex5:
I suspect that if mirage-firewall was updated to use the new ARP marshallers most of the code there would disappear. Not sure if I'd use this static ARP implementation, since mirage-firewall has its own data-structure with the mapping already.

djs55:
It looks like the docker/vpnkit implementation could go away and be replaced with this -- we'd just have to call `add_entry` a few times to add the entries and we'd be done... is that right?

hannesm:
while this serves well for testing purposes (and the use cases you mentioned), wouldn't a common case be to mix static and dynamic entries (which doesn't seem to be possible before or after this PR)?  (i.e. in a hostile network, statically `arp -s` record the mac address of the gateway, but record other hosts on the same broadcast domain dynamically)

yomimono:
Good point, @hannesm - I hadn't thought about the case where a user might want to static and dynamic entries but it's indeed reasonable, and probably more common than a user wanting a fully static ARP table.  Perhaps it would make more sense to provide a version which accepts dynamic updates when they don't conflict with static ones.

hannesm:
I'm not yet sure how the interface should be.  The current implementation has certainly some pitfalls (asking your gateway every 60 seconds for its mac address, DoS by flooding the arp table (which is not bound by memory)).  In some scenarios I'd appreciate a fully static ARP table (testing, controlled host including VMs), others should have fully dynamic one (the common mode), and some a static mixed with dynamic entries (it seems that got out of fashion, but I attended various conferences where printouts with the mac address of the gateway were available).  The underlying cache might be a size-restricted LRU cache or a Map ([why hash tables are a bad idea](http://www.securityweek.com/hash-table-collision-attacks-could-trigger-ddos-massive-scale)).

And then, either provide the three modules and let the builder of the stack choose their own poison or provide another way to configure the `accept_dynamic_entries` and `add/remove_static_entry`.  Another question would be whether to insert entries when we receive IPv4 frames (originating from the same network) or ARP requests.

Having said that, this PR is an improvement in some scenarios over the current state, and thus appreciated!

yomimono:
> The underlying cache might be a size-restricted LRU cache or a Map

An implementation which uses a reference to a Map instead of a hash table wouldn't be too difficult to plumb in, I think.  I haven't used [ocaml-lru-cache](https://github.com/zoggy/ocaml-lru-cache), but the interface seems to provide everything we'd need; perhaps someone is keen to try that out and and report back?

> either provide the three modules and let the builder of the stack choose their own poison 

I suspect this is the right way to go.  Perhaps an additional module type (`ARP_CONFIGURABLE` or something) which includes both the current `ARPV4` and also `add/remove_static_entry`? This would make the dynamic-with-static and fully-static cases interchangeable on the module level, but would disallow easily switching between fully-dynamic and dynamic-with-static/fully-static.

> Another question would be whether to insert entries when we receive IPv4 frames (originating from the same network) or ARP requests.

Is there a downside to this?  I recall having thought about and discussed this a bit some months ago, but I don't remember whether there was any particular reason not to do it.

talex5:
Could the user just supply a lookup function?

yomimono:
There's logic not just in lookups but also in updates -- the fully dynamic ARP cache takes this out of your hands entirely and hides all of it in the packet input function, but it's there.  The static and dynamic ARP in this patch don't have much custom beyond the lookup, update, and (in the case of dynamic ARP) expiration code.

hannesm:
I think there's no downside of learning dynamic ARP entries from IPv4/ARP requests (apart from the cache getting bigger -- thus some people might want to disable it).  Providing multiple modules sounds good to me, with the default having dynamic behaviour as now (but accepting static entries, using some bits of your static ARP module).

Numbers: FreeBSD uses 20 minutes for an ARP entry (we currently have 60 seconds); FreeBSD sends an ARP request if expiration time is approaching (we drop the entry and re-send a request upon demand (when the next packet needs to be send) -- some eager re-requesting (at least for the gateway) would be nice to have), maximum retries 5 (we 3).  In case a reply (or gratitious) is received which is already present in the cache, a warning is signalled (we just use the new entry).  Harvesting IPv4 frames is not done in FreeBSD (and are actual layering violations, why should the ARP handler know of IPv4 frame format).

talex5:
In the case of mirage-firewall, we want to be sure no-one can poison the ARP cache, so having an implementation with no ability to update state is very nice. But I'm quite happy having my own implementation.

yomimono:
I'm going to close this out.  Note there is a new implementation that came out of this discussion at https://github.com/hannesm/arp -- we might want to consider replacing the implementation in tcpip with this entirely.

hannesm:
note that nevertheless the ARP module type should be adjusted (to be able to add static entries, I added a comment to https://github.com/mirage/mirage/issues/557).  my idea would instead of "replacing" be to make it easier in the mirage tool to use another arp implementation (`tcpip.arpv4` is pretty hardcoded there)

PR #219, opened 2016-06-30
runs tests even when attempting to create pcap got a Unix exception

### LINKED ISSUE #201: Unit-tests pass if an exception is thrown

yomimono:
fixes #201 .

### DIFF
--- a/lib_test/test_iperf.ml
+++ b/lib_test/test_iperf.ml
@@ -126,6 +126,8 @@ module Test_iperf (B : Vnetif_backends.Backend) = struct
     Lwt.return_unit
 
   let iperf c s server_done_u flow =
+    (* debug is too much for us here *)
+    Logs.set_level ~all:true (Some Logs.Info);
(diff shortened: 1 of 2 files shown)


talex5:
LGTM (though there is some minor whitespace/indentation strangeness you might want to adjust)

yomimono:
Thanks for pointing out the indentation weirdness, @talex5 !

Issue #217, opened 2016-06-29
let users check checksums easily


yomimono:
For applicable `_unmarshal` modules, it should be easy to make decisions based on whether the checksum was valid.

yomimono:
Fixed via #277 .

PR #226, opened 2016-07-15
Fix TCP and UDP checksum brokenness

yomimono:
Fixes incorrect calculation of TCP and UDP layer checksums.  Thanks to @ijc25 for the error report.


hannesm:
may I ask since when this is broken?   was it broken in any released version of tcpip?

yomimono:
Absolutely you may!  This brokenness came in with https://github.com/mirage/mirage-tcpip/pull/205 .  There hasn't been a release since then.

yomimono:
(It wasn't caught in tests because our input functions don't verify checksums - see https://github.com/mirage/mirage-tcpip/issues/217 .)

ijc25:
I can confirm that this fixes the miraga-skeleton ethifv4 and static_website examples. Thanks!

PR #227, opened 2016-07-15
process incoming ARPs & construct GARPs per RFC 826.

### LINKED ISSUE #225: Gratuitous ARP is wrong
RFC 5227, section 3 (https://tools.ietf.org/html/rfc5227#section-3) says it should be a ARP request with spa = tpa (mirage-tcpip uses a ARP reply with tpa = broadcast).  Section 3 also makes it clear why an ARP reply is a bad idea.  Related to https://github.com/mato/docker-unikernel-runner/issues/1 (setting `arp_accept` (somewhere in proc on linux) to the bridge and sending an ARP request solves the issue (just verified with @mato)).

Discovered while reimplementing ARP (https://github.com/hannesm/arp - https://hannes.nqsb.io/Posts/ARP)

yomimono:
Per RFC 826, don't look at the `op` field of the packet before deciding
whether to use it to update the cache.  Also, construct gratuitous ARP
packets as Requests, not Replies, and set the destination host address
to the source host address.  This fixes #225.

### DIFF
--- a/lib/arpv4/arpv4.ml
+++ b/lib/arpv4/arpv4.ml
@@ -103,14 +103,12 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
   let input t frame =
     let open Arpv4_packet in
     MProf.Trace.label "arpv4.input";
-    Log.debug (fun f -> f "ARP packet received: %a" Cstruct.hexdump_pp frame);
     match Unmarshal.of_cstruct frame with
     | Result.Error s ->
-      Log.info (fun f -> f "Failed to parse arpv4 header: %a (buffer: %S)"
+      Log.debug (fun f -> f "Failed to parse arpv4 header: %a (buffer: %S)"
                    Unmarshal.pp_error s (Cstruct.to_string frame));
       Lwt.return_unit
     | Result.Ok arp ->
-      Logs.debug (fun f -> f "ARP packet received: %a" Arpv4_packet.pp arp);
       notify t arp.spa arp.sha; (* cache the sender's mapping. this will get GARPs too *)
       match arp.op with
       | Arpv4_wire.Reply -> Lwt.return_unit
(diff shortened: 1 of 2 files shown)


lib/arpv4/arpv4.ml
```diff
@@ -122,27 +123,30 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
...
   (* Send a gratuitous ARP for our IP addresses *)
   let output_garp t =
-    let tha = Macaddr.broadcast in
     let sha = Ethif.mac t.ethif in
+    let tha = sha in
     let tpa = Ipaddr.V4.any in
     Lwt_list.iter_s (fun spa ->
-        Log.info (fun f -> f "ARP: sending gratuitous from %a" Ipaddr.V4.pp_hum spa);
-        output t Arpv4_packet.({ op=Arpv4_wire.Reply; tha; sha; tpa; spa })
+        let arp = Arpv4_packet.({ op=Arpv4_wire.Request; tha; sha; tpa; spa }) in
```
hannesm: If I'm reading the spec correctly, the GARP should be an ARP request (which is good here) where `spa = tpa = your ip` (here `tpa` is still `Ipaddr.V4.any`).  The `tha` is not defined (common stacks use `00:00:00:00:00:00` for sending, and ignore it on receiving (`tcpdump` prints tha if it is not all zeroes).

yomimono:
@hannesm, you're pretty clearly passing a reading test I'm failing, so it'd be nice to get your Y/N.  Thanks!

lib/arpv4/arpv4.ml
```diff
@@ -73,21 +73,24 @@ module Make (Ethif : V1_LWT.ETHIF) (Clock : V1.CLOCK) (Time : V1_LWT.TIME) = str
...
-    let now = Clock.time () in
-    let expire = now +. arp_timeout in
-    try
-      match Hashtbl.find t.cache ip with
-      | Pending (_, w) ->
-        Hashtbl.replace t.cache ip (Confirmed (expire, mac));
-        Lwt.wakeup w (`Ok mac)
-      | Confirmed _ ->
+    Log.debug (fun f -> f "notifying: %a -> %s" Ipaddr.V4.pp_hum ip (Macaddr.to_string mac));
+    match Ipaddr.V4.is_multicast ip || (Ipaddr.V4.compare ip Ipaddr.V4.any = 0) with
```
hannesm: could you indicate why 0.0.0.0 and multicast addresses are filtered here? what about broadcast? (maybe point me to the RFC)

hannesm:
@yomimono the garp are now constructed as proposed in the RFC :+1:

lib/arpv4/arpv4.ml
yomimono: http://www.ietf.org/rfc/rfc894.txt pretty heavily implies that ARP is only reasonable to use with unicast IPs.  I agree that it would be good to also filter out broadcast but the current API makes that annoying; I'd like to include that in future improvements, after resolving https://github.com/mirage/mirage/issues/557 .

PR #228, opened 2016-07-18
cut off payload in buf before calculating the checksum

sevenEng:
cut off the contents in payload, as they are already included in `~payload` , otherwise `unsafe_fill` gives an incorrect checksum

### DIFF
--- a/lib/udp/udp_packet.ml
+++ b/lib/udp/udp_packet.ml
@@ -65,7 +65,8 @@ module Marshal = struct
       else Ok ((Cstruct.len payload) + sizeof_udp)
     in
     check_header_len () >>= check_overall_len >>= fun len ->
-    unsafe_fill ~pseudoheader ~payload t udp_buf len;
+    let buf = Cstruct.sub udp_buf 0 Udp_wire.sizeof_udp in
+    unsafe_fill ~pseudoheader ~payload t buf len;
     Ok ()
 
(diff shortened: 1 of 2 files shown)


lib/udp/udp_packet.ml
```diff
@@ -65,7 +65,8 @@ module Marshal = struct
       else Ok ((Cstruct.len payload) + sizeof_udp)
     in
     check_header_len () >>= check_overall_len >>= fun len ->
-    unsafe_fill ~pseudoheader ~payload t udp_buf len;
+    let buf = Cstruct.sub udp_buf 0 Udp_wire.sizeof_udp in
+    unsafe_fill ~pseudoheader ~payload t buf len;
```
yomimono: Thanks for the PR.  Rather than this solution, I'll have `unsafe_fill` limit the size of what it sends on -- that way potential future users of `unsafe_fill` won't be surprised, and the preparation for invocation in `into_cstruct` is a bit clearer.

yomimono:
Thanks for the PR. Rather than this solution, I'll have unsafe_fill limit the size of what it sends on -- that way potential future users of unsafe_fill won't be surprised, and the preparation for invocation in into_cstruct is a bit clearer.

yomimono:
@sevenEng , I've made a PR on top of your commit at https://github.com/mirage/mirage-tcpip/pull/229 - can you have a look at it?

PR #229, opened 2016-07-19
fix checksum calculation for into_cstruct in tcp and udp

yomimono:
Per @sevenEng 's commit, we incorrectly include the payload information in the checksum twice if `into_cstruct` is passed a buffer larger than the size of the header it expects to write.  Limit the size of the header buffer passed to `ones_complement_list` to avoid this.


sevenEng:
looks good to me :) this should make the checksum calculation more reliable then. 
should i do something with my previous PR?

sevenEng:
looks good to me :) this should make the checksum calculation more reliable then. 
should i do something with my previous PR?

yomimono:
I'll merge both :) thanks!

PR #231, opened 2016-08-16
Fix icmp echo replies

yomimono:
See https://github.com/mirage/mirage-net-solo5/issues/2 .  Fix buggy ICMP echo replies.


hannesm:
looks good to me.  a regression test would be great to have, though!

yomimono:
Thanks @hannesm ; there was a test that should've caught this but didn't, which I've fixed.

PR #235, opened 2016-08-26
Don't rely on Hashtbl.filter_map_inplace, breaks 4.02.3.

haesbaert:
This was only introduced in 4.03.0.

I couldn't really test it.


yomimono:
Ack, my apologies; @hannesm even mentioned this in an inline comment, but I forgot to back it out before merging.

yomimono:
I had to make a small correction to this; I've opened #236 with the followon.

PR #238, opened 2016-08-30
Fix ip parsing when there are options on the datagram

polvorin:
Must calculate the options' length.

The IP parsing code was crashing trying to parse some packets that contained IP options, because the option length was incorrect.


yomimono:
Thanks for this patch, @polvorin , and for the tests.

PR #241, opened 2016-09-08
add duration library to .merlin

yomimono:


yomimono:
Tests are failing because of `mirage-www`, which has an open issue in that repository, so I'll merge this.

PR #243, opened 2016-09-09
Support IP fragmentation

### LINKED ISSUE #237: ICMP packets larger than interface MTU are not correctly processed
To reproduce, pick any unikernel with a network stack. Send it an ICMP echo packet larger than the interface MTU, e.g. `ping -s 2000 -c 1 10.0.0.2`.
Unikernel output:

```
2016-08-29 15:17:50 +02:00: INF [icmpv4] ICMP: error parsing message from 10.0.0.1: unrecognized ICMPv4 type
```

`ping` output:

```
PING 10.0.0.2 (10.0.0.2) 2000(2028) bytes of data.
1480 bytes from 10.0.0.2: icmp_seq=1 ttl=38 (truncated)

--- 10.0.0.2 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 0.231/0.231/0.231/0.000 ms
```

Reproducible on (at least) `unix`, `virtio` and `ukvm` targets so looks like an ICMP issue.

polvorin:
Hello,  this fixes #237   , allowing the IP stack to reassemble fragmented packets, as well as fragment outgoing ones larger than the network mtu.

Notes:
-  Uses a fixed mtu of 1500, as the network layer currently doesn't give a way to ask it.  Should be good to add that to the ethif drivers
-  I included on the commit some oasis generated files. Not sure if they are ok. (need to change it to include IPv4 dependency on 'duration' module)
- The Ipv4 functor now requires a Time parameter as well,so need to update the scripts that do the network stacks config for mirage.
- Ipv4 write/writev should have an optional argument for setting the 'don't fragment' flag, that can be used for TCP Path MTU Discovery.


yomimono:
Thanks a lot for this patch, @polvorin .  I haven't had a chance to look at it in depth, but here are some thoughts on the notes you called out in the description:

>  Uses a fixed mtu of 1500, as the network layer currently doesn't give a way to ask it. Should be good to add that to the ethif drivers

A way to set this for the ethif layer would also be good -- we might receive this in a DHCP lease, but we currently have no way to honor it.

>  I included on the commit some oasis generated files. Not sure if they are ok. (need to change it to include IPv4 dependency on 'duration' module)

That's fine for the moment.  I'm hoping to port `mirage-tcpip` to `topkg` to avoid including all of these autogenerated files in the future.

> The Ipv4 functor now requires a Time parameter as well,so need to update the scripts that do the network stacks config for mirage.

Are you interested in putting a PR together for https://github.com/mirage/mirage ?  There are a few examples from similar additions that I can point you at which may be helpful, if so.  If not, another contributor can probably set this up, but it may take a little while.

> Ipv4 write/writev should have an optional argument for setting the 'don't fragment' flag, that can be used for TCP Path MTU Discovery.

I'd like to mull this over a bit.  An optional argument is an easy way to allow the caller to specify the behavior, but it might be quite difficult for the user to inject this request into the stack.

yomimono:
@hannesm , @polvorin and I discussed the fragmentation behavior in IRC a few days ago.  @polvorin , please feel free to ping with any more comments or to let us know you'd like another look.

polvorin:
Hi @yomimono ,  I did not had time to work on this these days.   I've modifying the behaviour acording to discussion regarding possible attack vectors (last commit),  and improved the tests a little bit, but that's it,  I haven't  advanced on the broad pull request to make this change all the way up across the different repos that would need to adapt.   
By the way,  as changing those interfaces is laborious,  perhaps we should first settle on how ipv4 should let the upper layers specify the 'don't fragment'  flag. I think that would be required as soon as anyone try to implement path mtu discovery.

avsm:
Although this was closed due to conflicting heavily, the feature seems like it could still be integrated. Should we open an issue to track it and do as @polvorin suggested with respect to settling on how ipv4 should let upper layers specify the DF flag?  @samoht @yomimono

yomimono:
Yes, that would be a good idea.

PR #239, opened 2016-09-06
Fix timer recovery bug update

balrajsingh:

PR #244, opened 2016-09-12
Fix timer recovery bug with test

yomimono:
Tests for incorrect window calculation when entering fast-retransmit, and Incorporates @balrajsingh 's fix in #239 .


jmitchell:
Thanks @balrajsingh and @yomimono!

PR #246, opened 2016-09-20
Test improvements

MagnusS:
- Fixes a race in the pcap recording code, where the last packets in the trace could be lost if the file was closed too quickly after the last packets were sent (requires mirage-vnetif 0.2.0)
- Adds a new backend to only drop packets with no/small payload, as suggested by @balrajsingh. The backend is used by a new iperf test and a slower connect x 100 times test 
- Adds a per-test timeout to the iperf tests


yomimono:
This looks good to me.  I've kicked the failing test, which looked transient to me.

PR #248, opened 2016-09-22
New backend to trigger fast retransmit in iperf tests

MagnusS:
Adds a new backend that drops packets for 1 sec after 1mb transferred, as suggested by @balrajsingh . This should trigger a fast retransmit and the test should complete relatively quickly. If a fast retransmit is not triggered, the iperf test will time out (total time to complete will be 6-7 minutes in the broken case).

This test should time out with mirage-tcpip 2.8.0, but work in 2.8.1 after the recovery bugs were fixed in #239.

The PR has some conflicts with #246 - I'll rebase one of them depending on which one is merged first .

### DIFF
--- a/lib_test/test_iperf.ml
+++ b/lib_test/test_iperf.ml
@@ -98,7 +98,7 @@ module Test_iperf (B : Vnetif_backends.Backend) = struct
 
   let iperfclient s amt dest_ip dport =
     let iperftx flow =
-    Logs.info (fun f -> f  "Iperf client: Made connection to server.");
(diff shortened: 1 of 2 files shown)


yomimono:
Something strange seems to be happening with your indenter :)

Why do the tests take 6-7 minutes in the broken case, when the timeout is set to `120.0`?

MagnusS:
@yomimono it would take that much time if the timeout hadn't been there

Hm it looks like may I have to fix my editor setup... re-indented and rebased

Issue #249, opened 2016-09-27
Packets only contain 536 bytes or less


MagnusS:
From iperf pcap (iperf test 0):
```
$ tshark -nr tests/pcap/tcp_iperf_two_stacks_basic_10000000.pcap |more
 ...
  5   0.106235   10.0.0.101 → 10.0.0.100   TCP 62 10863→5001 [SYN] Seq=0 Win=5840 Len=0
  6   0.106436   10.0.0.100 → 10.0.0.101   TCP 58 5001→10863 [SYN, ACK] Seq=0 Ack=1 Win=65535 Len=0
  7   0.106554   10.0.0.101 → 10.0.0.100   TCP 54 10863→5001 [ACK] Seq=1 Ack=1 Win=65535 Len=0
  8   0.106791   10.0.0.101 → 10.0.0.100   TCP 590 10863→5001 [PSH, ACK] Seq=1 Ack=1 Win=65535 Len=536
  9   0.106819   10.0.0.100 → 10.0.0.101   TCP 54 5001→10863 [ACK] Seq=1 Ack=537 Win=65535 Len=0
 10   0.106863   10.0.0.101 → 10.0.0.100   TCP 590 10863→5001 [PSH, ACK] Seq=537 Ack=1 Win=65535 Len=536
 11   0.106879   10.0.0.100 → 10.0.0.101   TCP 54 5001→10863 [ACK] Seq=1 Ack=1073 Win=65535 Len=0
 12   0.106905   10.0.0.101 → 10.0.0.100   TCP 442 10863→5001 [PSH, ACK] Seq=1073 Ack=1 Win=65535 Len=388
 13   0.106920   10.0.0.100 → 10.0.0.101   TCP 54 5001→10863 [ACK] Seq=1 Ack=1461 Win=65535 Len=0
[etc]
```

From debugging with @balrajsingh it looks like the MSS option is not included in the SYN and it falls back to 536 bytes.

This seems to be a regression in master, as 2.8.1 has the correct packet length.

yomimono:
Fixed by #249.

PR #250, opened 2016-09-27
Fix for #249 - removed redundant memset that was zeroing out all options

balrajsingh:
Redundant since every header field must be set and is being set (or memoized).

### DIFF
--- a/lib/tcp/tcp_packet.ml
+++ b/lib/tcp/tcp_packet.ml
@@ -73,7 +73,6 @@ module Marshal = struct
   let unsafe_fill ~pseudoheader ~payload t buf options_len =
     let data_off = (sizeof_tcp / 4) + (options_len / 4) in
     let buf = Cstruct.sub buf 0 (data_off * 4) in
-    Cstruct.memset buf 0; (* autopad *)
     set_tcp_src_port buf t.src_port;
     set_tcp_dst_port buf t.dst_port;
     set_tcp_sequence buf (Sequence.to_int32 t.sequence);


lib/tcp/tcp_packet.ml
```diff
@@ -73,7 +73,6 @@ module Marshal = struct
   let unsafe_fill ~pseudoheader ~payload t buf options_len =
     let data_off = (sizeof_tcp / 4) + (options_len / 4) in
     let buf = Cstruct.sub buf 0 (data_off * 4) in
-    Cstruct.memset buf 0; (* autopad *)
```
avsm: looks like this got added in 4f3acd9968e12843d5134b4d28b around the time that Cstruct zeroing policy also got stricter, hence the confusion. We should add a test case for checking this doesnt regress as well - I can do that tomorrow if you dont have one already @balrajsingh

balrajsingh: Don't see how zeroing policy change could affect it since as long as it was doing a memset it would always blow away the options that had been set before this call.  Of course if it was not doing a memset then it would explain the change.  And also confirm that this line is not required.  I don't have a test, if you can think of a test for this please do add.  Thx.

yomimono: The `(* autopad *)` comment is a bit cryptic -- what I meant here was "we may have to include some padding in the TCP options, so make sure this padding is set to `\000` as is appropriate."  @avsm is right that this patch was merged shortly before `Cstruct.create` was changed to zero memory by default; `tcpip` (and indeed many `Cstruct` users) could use another look to see whether there are unnecessary `memset` calls generally.

yomimono: @balrajsingh , am I correct in presuming that this PR is the result of some erroneous behavior on the part of the TCP stack in some existing application?  If you point me at it, I can try putting a test together for you; it seems to me that you were motivated by noticing incorrect writes for options, so we could write a simple unit test for `Marshal.into_cstruct` which asserts that the returned buffer includes nonzero values representing the options passed.

MagnusS: @yomimono it can be reproduced with the iperf tests in the test suite, see #249

yomimono: Thanks @magnuss, I missed the link here.

yomimono: `test_tcp_options` was testing the results of `Options.marshal`, but not the presence of the options field in a generated packet.  I've added a test which fails on current master in https://github.com/yomimono/mirage-tcpip/commit/abbdbe9c5ad1dbb161143981f9a4534d2c2cbec1 .

avsm: This lgtm to merge along with @yomimono's test

balrajsingh: Thanks @yomimono, @avsm, the test looks good to me too.  Was just discussing with @MagnusS - another test he is thinking of adding is to examine a small packet capture to see if behaves as expected - eg, in this case check that the SYN pkts have MSS options.

Issue #207, opened 2016-05-24
Move DHCP client code to a separate repository.


yomimono:
Now that robust support for DHCP is available in [charrua-core](https://github.com/haesbaert/charrua-core), an independent client that relies on that library is probably a better choice than a reimplementation here.  The `mirage` front-end tool can be responsible for instantiating a network stack to invoke `dhcp` on top of.

Issue #168, opened 2015-08-24
The `Dhcp` functor is applied inside `Tcpip_stack_direct.Make`


Drup:
See https://github.com/mirage/mirage-tcpip/blob/2c820a1c211ec63b6e8b60087799a499ca22e16e/lib/tcpip_stack_direct.ml#L54

If we follow the logic, this should be an argument.

Drup:
Actually, I went and look, and there is no `.mli` for the dhcp part, and it seems in general that this part didn't got a lot of love.

avsm:
In general it would be good to make the Dhcp stack a client of the Tcpip stack. The plan is to use  https://github.com/haesbaert/hdhcp as the code base for this rather than the rather adhoc one thats currently in the mirage-tcpip codebase. (/cc @haesbaert)

mor1:
There is also the code that Alistair (UROP student) has been working on. At
some point I believe he was in touch with Christiano but not sure where
that led. He seemed to have made some progress on producing a dhcpd at
least, and was (I believe) due to be talking with @yomimono about applying
some Irmin goodness to it. (I don't think he'd yet extracted it from
mirage-tcpip though.)

On 25 August 2015 at 13:30, Anil Madhavapeddy <notifications@github.com>
wrote:

> In general it would be good to make the Dhcp stack a client of the Tcpip
> stack. The plan is to use https://github.com/haesbaert/hdhcp as the code
> base for this rather than the rather adhoc one thats currently in the
> mirage-tcpip codebase. (/cc @haesbaert <https://github.com/haesbaert>)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mirage/mirage-tcpip/issues/168#issuecomment-134570460>
> .
>



-- 
Richard Mortier
mort@cantab.net

haesbaert:
Not really, no one contacted me, wasn't aware there was another
implementation going on.

I have released the library needed for hdhcp on opam this weekend and hdhcp
seems to be functional, I've started to work on converting it to mirage.

mor1:
@haesbaert Drat, sorry about that. My bad :( /cc @alistairfisher
Alistair has a Mirage DHCP server working, but based off the DHCP code that was here in mirage-tcpip so less complete than yours. After discussion with @yomimono, he's also started to work on applying Irmin to hold the state of the server. 

How far have you got with the Mirage server version so far -- perhaps worth the two of you joining forces?

haesbaert:
[...]

PR #260, opened 2016-11-02
depend on charrua-client for dhcp client

yomimono:
Remove the custom DHCP client and instead use `charrua-client` for DHCP configuration of ipv4 addresses.  Requires https://github.com/mirage/mirage/pull/643 and [charrua-client](https://github.com/yomimono/charrua-client), which is not in opam but available in the mirage-dev fork on my repository under the `normalize-ipv4` branch (which represents a consistent universe for this set of changes).


yomimono:
Fixes #207 and #168

hannesm:
I do like this.  But somehow an IP stack itself should not care about calling out to DHCP, or should it?  (read: I would like to get rid of the charrua-client dependency here).

I'd be happy if charrua-client was a functoria/mirage-level device, and gets an UDP implementation, outputs at some point maybe a DHCP lease, which is then used to configure a full TCP/IP stack.  And this is the only constructor of a TCP/IP stack: you have to give it an IP/netmask/gateway (of course, the dummy udp stack for the DHCP client uses 0.0.0.0 etc.).

At bootup, if your web server needs a full stack, it has to wait until after DHCP figured out an IP address, and created a STACK with that information (same is true for the syslog reporter -- there's no way in a DHCP setting to get the syslog messages (apart from logging to the broadcast address) until after an IP has been allocated.

yomimono:
@hannesm : if I understand right, you propose something like:

`dhcp_client : udp @-> dhcp_lease`
`dhcp_tcpip_stack : `dhcp_lease @-> network @-> stack4` ?

> At bootup, if your web server needs a full stack, it has to wait until after DHCP figured out an IP address, and created a STACK with that information 

Yes.  IMO that's fine; if you have enough information to operate without a DHCP lease, why did you ask for one?

hannesm:
I guess what I'm asking for is to move the DHCP logic out of the tcpip library (optimally into the dhcp client library), the only leftover bits I can see are:

``` OCaml
module Make(Dhcp_client : DHCP_CLIENT) (Ethif : ETHIF)(Arp : ARP) = struct
  (* for now, just wrap a static ipv4 *)
  module I = Static_ipv4.Make(Ethif)(Arp)
  include I
  let connect dhcp ethif arp =
    Lwt_stream.last_new dhcp >>= fun (config : ipv4_config) ->
    I.connect ~ip:config.address ~network:config.network ~gateway:config.gateway ethif arp
end
```

If I look into a TCP/IP stack and just assume an IP and netmask, instead of thinking where they came from, I've to think about less.  But this might as well be a patch on top of this PR.

yomimono:
Oh, I see!  Yes, now that you mention it, `dhcp_ipv4.ml` could move into charrua-client indeed.  In that case it wouldn't be necessary to rename Ipv4 to Static_ipv4 either, although perhaps that's desirable still.

hannesm:
if there's only `Ipv4`, no need to rename it.  I like the `_common` though for code independent of the applied functor...

yomimono:
I'm going to leave it `Static_` after a bit of further thought.  Feel free to revert or change if you dislike it, or make a more vehement argument. :)

mato:
Something is screwed up after this (and related) merges. Testing on Solo5 (ukvm and virtio) I can observe the following behaviour (all with `static_website_tls`):
- using the default configuration, it looks like the outgoing (TX) path works as I see the initial GARP packet from the unikernel, however it does not appear to respond to any incoming (RX) packets.
- configured with `--dhcp=true` it successfully gets a lease (so the dhcp conversation succeeds), but then promptly continues the same behaviour as in the static IP case (not responding to anything, including ARP)

Related: The new DHCP client is somewhat silent, the only message I can see logged about what lease it has obtained is: `2016-11-04 11:50:29 -00:00: INF [udp] UDP interface connected on 10.0.0.20`. This is a bit cryptic, would be nicer to have something more informative.

yomimono:
Are you calling Stack.listen ? Previously the stack would do it for you in a way not recommended by Lwt.

On November 4, 2016 6:56:53 AM CDT, Martin Lucina <notifications@github.com> wrote:
>Something is screwed up after this (and related) merges. Testing on
>Solo5 (ukvm and virtio) I can observe the following behaviour (all with
>`static_website_tls`):
>- using the default configuration, it looks like the outgoing (TX) path
>works as I see the initial GARP packet from the unikernel, however it
>does not appear to respond to any incoming (RX) packets.
>- configured with `--dhcp=true` it successfully gets a lease (so the
>dhcp conversation succeeds), but then promptly continues the same
>behaviour as in the static IP case (not responding to anything,
>including ARP)
>
>Related: The new DHCP client is somewhat silent, the only message I can
>see logged about what lease it has obtained is: `2016-11-04 11:50:29
>-00:00: INF [udp] UDP interface connected on 10.0.0.20`. This is a bit
>cryptic, would be nicer to have something more informative.

composed on a touchscreen keyboard; please forgive typos and brevity

mato:
Probably not, I'm just using the existing `static_website_tls` from mirage/mirage-skeleton#mirage-dev.

yomimono:
I'll push fixes shortly. I cleaned this up where I saw it in the tcpip tests but completely blanked on the higher-level examples in skeleton. Thanks!

On November 4, 2016 7:43:06 AM CDT, Martin Lucina <notifications@github.com> wrote:
>Probably not, I'm just using the existing `static_website_tls` from
>mirage/mirage-skeleton#mirage-dev.

composed on a touchscreen keyboard; please forgive typos and brevity

yomimono:
@mato `static_website_tls`, `static_website`, `conduit_server`, and `http_fetch` all do their stack management via `mirage-http` or `conduit`, so it's taking me a little longer to fix them.  The `network` example might work just as well for your purposes and shouldn't have this problem, if you're blocked on this today.

PR #263, opened 2016-11-08
Restore the previous call to `listen` in STACK.connect, and document it.

yomimono:
Requiring each caller to invoke `STACK.listen` is currently rather annoying, because there is no interface for any module that's passed a `STACK.t` to know whether what passed the `STACK.t` to it has called `listen` or not.  Restore the behavior that was removed in https://github.com/mirage/mirage-tcpip/commit/cccd214dcb5d3e8e17e07982317a02555780e704#diff-0089e4d851e40832017a323e072da706L180 , and attempt to do a better job of documenting it.  (A companion PR in the `mirage` repository will add more docstrings to the module type description for `V1.STACKV4`.)

### DIFF
--- a/lib/tcpip_stack_direct.ml
+++ b/lib/tcpip_stack_direct.ml
@@ -104,7 +104,7 @@ struct
     with Not_found -> None
 
   let listen t =
-    Logs.debug (fun f -> f "Initializing listener for stack %a" pp t);
+    Logs.debug (fun f -> f "Establishing or updating listener for stack %a" pp t);
     Netif.listen t.netif (
       Ethif.input
         ~arpv4:(Arpv4.input t.arpv4)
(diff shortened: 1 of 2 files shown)


yomimono:
I'm not too keen on this behavior, and I'd like to figure out something a little less surprising -- it seems most users aren't aware of what happens with multiple calls to `listen`.  If you have ideas, there is some existing conversation in https://github.com/mirage/mirage/issues/426 that could be continued there.

Since this PR restores functionality for examples in `mirage-skeleton` that are currently broken, I'd prefer to keep discussion here focused on the bugfix and have larger discussions on the API design on the issue linked above.

hannesm:
I suspect you'll want to revert https://github.com/mirage/mirage-skeleton/pull/191 once this is merged.

yomimono:
Indeed, thanks @hannesm.

PR #265, opened 2016-11-14
give only options part when trying to unmarshal a tcp packet

sevenEng:
or the `unmarshal` will try to interpret both options part and TCP payload part, which gives a wrong options field (very often larger than 40 bytes)

### DIFF
--- a/lib/tcp/tcp_packet.ml
+++ b/lib/tcp/tcp_packet.ml
@@ -41,7 +41,8 @@ module Unmarshal = struct
     in
     let options data_offset pkt =
       if data_offset > 20 then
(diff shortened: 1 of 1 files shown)


yomimono:
Good catch -- the test in https://github.com/mirage/mirage-tcpip/blob/master/lib_test/test_tcp_options.ml#L63 checks to see that parsing ends at EOF, but not that the parser does the right thing when EOF isn't properly signaled.  Do you by chance have a test handy that we could put into `test_tcp_options`.

sevenEng:
Sorry, I was doing some rebasing of the PR onto the latest two commits and deleted my repo's original branch

I'll reopen one after adding some useful test bits in `test_tcp_options.ml`

PR #274, opened 2016-12-14
requires mirage-types-lwt and bump version constraint to 3.0.0

hannesm:


hannesm:
build failure which should be fixed by this at e.g. https://travis-ci.org/hannesm/arp/jobs/183963745

PR #277, opened 2016-12-21
add verify_transport_checksum to ipv4_packet.unmarshal

yomimono:
Add a function `verify_transport_checksum` to ipv4_packet.unmarshal.  As
the function needs to use Tcp_wire and Udp_wire, include the UDP and
TCP sublibraries when building Ipv4.

Also include some tests against pre-created packets with good checksums.

/cc @haesbaert , as this is intended for use to replace the checksum validation that I removed a little while ago in charrua-core.


_oasis
```diff
@@ -58,7 +58,9 @@ Library ipv4
   Modules:            Static_ipv4, Ipv4_wire, Ipv4_packet
   InternalModules:    Ipv4_common, Routing
   BuildDepends:       logs,io-page,mirage-types,ipaddr,cstruct.ppx,lwt,result,rresult,
-                      tcpip,tcpip.ethif,mirage-runtime
+                      mirage-runtime,
```
samoht: do we still need `mirage-runtime` in there?

yomimono: No, #283 removes it.

yomimono: (It is needed with *this* patchset.)

PR #281, opened 2016-12-27
Adapt to the new mirage-net package

samoht:


samoht:
Tests are fixed by https://github.com/MagnusS/mirage-vnetif/pull/12

Issue #257, opened 2016-10-06
examples directory


hannesm:
is not automatically built by anything... either hook it up to CI; or move to mirage-skeleton?

avsm:
I'm building it as part of the topkg branch

yomimono:
@avsm, I heard a rumor that you're dropping your `topkg` branch.  Was there something unsupported or difficult about the package, or did I just churn it too much while you were trying to make changes?

PR #284, opened 2017-01-06
refer to mirage module types modules by new name

yomimono:
V1 is now Mirage_types, and V1_LWT is now Mirage_types_lwt, as of MirageOS version 3.0.0.

### DIFF
--- a/examples/services.ml
+++ b/examples/services.ml
@@ -1,7 +1,7 @@
 open Lwt
-open V1_LWT
+open Mirage_types_lwt
 
-module Main (C: V1_LWT.CONSOLE) (S: V1_LWT.STACKV4) = struct
+module Main (C: Mirage_types_lwt.CONSOLE) (S: Mirage_types_lwt.STACKV4) = struct
   let report_and_close c flow message =
     C.log c message;
     S.TCPV4.close flow


examples/services.ml
```diff
@@ -1,7 +1,7 @@
 open Lwt
-open V1_LWT
+open Mirage_types_lwt
```
samoht: Not sure about that one. Should we try to use `Mirage_console_lwt.S` and `Mirage_stack.V4` there? I am not sure to remember what these examples are supposed to show anyway, so please do what you think is best :-)

yomimono: I'm about to automatically generate a *lot* of PRs just like this one, so if you feel strongly that we should replace the references to V1 and V1_LWT with specific references to the module type from the package where it lives, now's the time to say so.

samoht: I think that would be better, but this can be done that progressively later on (e.g. it is not a blocker for me for the release). Obviously it is better to have something working first :-)

yomimono:
I shouldn't have merged this yet as the corresponding mirage/mirage PR is still open, but as noted in #257 nothing builds this automatically at the moment so my temporary lack of attention wasn't caught by a robot.

PR #285, opened 2017-01-12
update CHANGES: mirage-{protocols,stacks}, not types

yomimono:


yomimono:
Tests won't pass until https://github.com/mirage/mirage-www/pull/510 or something like it is merged, so please don't restart them as it just takes up test slots to report a known error.  I will re-run tests here once mirage-www is building.

PR #288, opened 2017-01-22
Compute TCP MSS from IP, ethif MTU

djs55:
Previously we hardcoded the value `1460` in various places where `1460` is `1500` - the size of the ethernet header - the size of an IPv4 header. This PR adds an accessor function `mtu: t -> int` to the ethernet interface and both IPv4 and IPv6, and then uses this to compute a TCP MSS.

Hopefully this is a small step on the way towards adding things like jumbo frame support.

This requires https://github.com/mirage/mirage-protocols/pull/4


yomimono:
I keep meaning to rebase this, thanks for being so patient - there's no other reason not to merge it.  I'll try to make sure to get it in today.

djs55:
Thanks!

PR #295, opened 2017-03-02
Avoid float in RTO calculations

mattgray:
Now we have `MCLOCK` and `int64` , we don't want to go back and forth between float when we calculate smoothed RTT, RTT variance and the retransmission timeout. This patch removes float in those local calculations. Written by @olleolleolle and me in Marrakech


yomimono:
Thank you very much @mattgray and @olleolleolle for this excellent PR, and in particular for the attached tests.

avsm:
This is so awesome!

hannesm:
Great work, thanks so much!

PR #299, opened 2017-03-03
reintroduce random into IPv6 reachable time computation

hannesm:
together with @olleolleolle we passed a Mirage_random.C to IPv6, and used it in
NDPv6 (using a higher-order function, no functors)

this depends on https://github.com/mirage/mirage/pull/801


hannesm:
unclear to me atm whether there is a socket tcpv6 stack which shares mirage IPv6 API (esp. connect) and needs updating...

yomimono:
Thanks for this PR.  I think it needs a bit more attention:

```
File "lib/ipv6/ipv6.mli", line 18, characters 17-32:
Error: Unbound module Mirage_random
Command exited with code 2.
```

olleolleolle:
(Short) the lib/META seems to be missing the `mirage-random` dependency in some of the "targets":

https://github.com/mirage/mirage-tcpip/blob/210d25ac5ea60cb863f324379aed9e2e045a42fb/lib/META#L69

hannesm:
updated: fixed `handle` in Ndpv6 to contain `~random`, and pass this on to `handle_ra`; also add `mirage-random` and `randomconv` dependency to `ipv6` sublibrary in `_oasis`.

this compiles fine now locally.

PR #300, opened 2017-03-04
Test for UDP on IPv6

mattgray:
Let's have some tests for IPv6. I will try to add further tests (TCP etc) in future.


mattgray:
Cool @yomimono ! I have a nearly completed TCP on IPv6 test, I plan to try and finish that this evening, so if you like I can fix up the random stuff then and give you a mergeable PR tomorrow. Up to you though.

PR #301, opened 2017-03-05
Fix various parsing bugs in tcpip

talex5:
They were checking that the buffer was big enough for the headers and
the payload, but they don't write the payload and there's no reason to
require it to be written to the same buffer.

Also, change the IPv4 marshal functions to take the payload length
rather than the payload itself, so that the caller can provide the
payload in multiple buffers. This also makes it clear that the function
does not actually write the payload.


talex5:
I've found a new more bugs and pushed fixes and tests for them to this branch too:

- The IP equal function didn't use Cstruct.equal.
- The IP header length calculation was incorrect when IP options were used.
- The IP parser didn't check that the buffer was big enough to contain the claimed payload length.
- TCP make_cstruct didn't include the options in the TCP checksum.

talex5:
Also, it tried to interpret the TCP payload as additional options, so pushed a fix for that.

yomimono:
Thanks for the fixes and especially the tests; this looks great.

PR #302, opened 2017-03-06
make some better _packet equals functions

yomimono:
ipv4_packet is fixed in #301.


talex5:
Maybe we should use [ppx_deriving](https://github.com/whitequark/ppx_deriving) to make this less fragile?

yomimono:
Without equality functions for types made via `cenum` attempts to use `[@@deriving eq]` in the `_packet` modules get pretty ugly pretty quick, compounded by the lack of equality for `Ipaddr.V4.t` and `Ipaddr.V6.t` and `Macaddr.t`.

PR #309, opened 2017-04-21
unbreak TCP

hannesm:
a recently introduced change (#288) breaks the TCP part of this library by sending out segments which are bigger than the mtu.  this patch reverts parts of that patch (sorry, I don't have energy or time to dig deeper how to 'properly' fix it).  I discovered the breakage when updating several unikernels to use tcpip-3.x, and run into assertion failures (https://github.com/Solo5/solo5/blob/v0.2.2/kernel/virtio/virtio_net.c#L123) (but only after some hours when someone with a strange tcpip stack (read: window size above 1460) came along).  I use this patched tcpip since ~1 week without any assertion failures.

### DIFF
--- a/lib/tcp/window.ml
+++ b/lib/tcp/window.ml
@@ -59,6 +59,7 @@ type t = {
 let count_ackd_segs = MProf.Counter.make ~name:"tcp-ackd-segs"
 
 let default_mss = 536
+let max_mss = 1460
 
 (* To string for debugging *)
 let pp fmt t =
@@ -77,7 +78,7 @@ let t ~rx_wnd_scale ~tx_wnd_scale ~rx_wnd ~tx_wnd ~rx_isn ~tx_mss ~tx_isn =
   let rx_nxt = Sequence.incr rx_isn in
   let rx_nxt_inseq = Sequence.incr rx_isn in
   (* TODO: improve this sanity check of tx_mss *)
-  let tx_mss = match tx_mss with |None -> default_mss |Some mss -> mss in
+  let tx_mss = match tx_mss with |None -> default_mss |Some mss -> min mss max_mss in
   let snd_una = tx_nxt in
   let fast_rec_th = tx_nxt in
   let ack_serviced = true in


djs55:
Do you happen to have a trace of the bad packets? Were the packets much bigger than the MTU or where they suspiciously close to the MTU (perhaps caused by forgetting to account for an IP or TCP header length)?

hannesm:
sorry, I don't have any trace.

I captured the _sending_ side (thus, a hexdump in virtio_net), and discovered packet length between 2955 and 3908 bytes (which contains the complete data the unikernel was supposed to send out).  there are (on the sending side) no fancy ip or tcp options present... but I don't have a full handshake.

I actually believe that a `max` here at this point is truely what we want - what was the motivation to remove it?

djs55:
I think I assumed the value was capped elsewhere. Obviously not! :)

I agree that we shouldn't send fragments larger than the MTU, but since ethernet jumbo frames are about ~9000 so imposing a limit of 1460 here seems wrong. I think we need to ask the ethif what the MTU is (1500 or 9000 or whatever) and use that instead of the constant?

hannesm:
Maybe.  I've no idea how to setup jumbo frames and test them (I cannot find any tests!?), and was mainly interested in getting my unikernels working again.  The commit above is a partial revert of the commit which broke TCP for me.

avsm:
@djs55 a normal tcp stack probably doesnt need the check as it would IP fragment, but we dont support that.  It would also do path MTU discovery which we also dont do.

perhaps we should add a check at the IP transmit layer to drop packets larger than the MTU with a warning logged...

hannesm:
@avsm is your suggestion serious?  You now want to push onto the user of TCP the invariant to never call write with a byte buffer bigger than the MTU?  sounds like a very error-prone design, in addition to breaking all existing users.

avsm:
That wasn't my suggestion -- the TCP stack should clearly segment according to the MSS and respect the MTU as part of the calculation, but it has no way of knowing what the _path_ MTU is and so might always transmit something that results in fragmentation. There is also a race between the interface MTU being changed and us transmitting.  An IPv4-level check to avoid transmitting > MTU would catch all of this as an additional sanity check and making finding similar issues easier.  I would prefer to avoid implementing IPv4 fragmentation in the stack if we can avoid it at all.

balrajsingh:
Some MTU related points (hope I remember them correctly):
- TCP should always know the MTU of it's own stack and should break writes bigger than the MTU into multiple segments.  Mirage TCP does this already and I think it does it correctly.  
- If along the path there is an interface with MTU smaller than the segment then that router should apply IP fragmentation to the segment.
- Path MTU discovery is definitely optional.  The only reasons for it are: if an errant router does not do IP frag when it is required to, and to avoid the cost of reassembling IP fragments at the dest.

So the main thing is that TCP should know the MTU of the interface that the pkt will be sent out on.  The application above TCP should be able to write any size - the amount written would be constrained by other practical matters but not the MTU.  In fact the application should not even know the MTU.  TCP should just break up the write into MTU sized segments. 

And it's definitely an error if an IP pkt is already fragmented when it leaves the source stack.

Mirage would only need to do IP frag if it is forwarding packets.  I don't know if the IP layer in Mirage does reassembly.  Implementing IP frag is very easy, but reassembly is a pain.

yomimono:
> I don't know if the IP layer in Mirage does reassembly

It doesn't.  We have an outstanding PR to implement this at #243 which could use some attention from someone who has free time; I got distracted from it during the MirageOS 3 release and haven't properly revisited it.

yomimono:
I will put some time into putting together some tests for this bit of code (it should be possible to test with `mirage-vnetif`).

avsm:
Should we merge this quickly until we fix the underlying issue properly @djs55?  

Sounds like `Ethif.mtu` querying + subtracting the pseudoheader is the way to fix it properly, but this might require figuring out what the IPv6 header size is...

yomimono:
The `Ip` module type has `mtu` for this purpose.

avsm:
Perfect!

djs55:
So rather than the constant `1460` should the `max_mss` be `Ip.mtu - <tcp header size>` (where normally ethernet MTU is 1500, IP header is 20 and TCP header is 20 so normally `max_mss=1500-20-20=1460`)?

djs55:
FYI on Xen it looks [mirage-net-xen](https://github.com/mirage/mirage-net-xen/blob/af58c2a21ea58dabaa6085103121977d60badf54/lib/frontend.ml#L285) calls [Cstruct.fillv](https://github.com/mirage/ocaml-cstruct/blob/a13beb27e30a83e6ea2d7e321cc0cd61e7bd859c/lib/cstruct.ml#L324) which only copies as much data into the destination as will fit. So I suspect we're emitting truncated ethernet frames.

talex5:
@djs55 I think that's only called from https://github.com/mirage/mirage-net-xen/blob/af58c2a21ea58dabaa6085103121977d60badf54/lib/frontend.ml#L319, which will call it as many times as needed.

Maybe you can deadlock it by sending a request so big that the ring doesn't have enough slots to list all the data pages, depending on what `Lwt_ring.Front.wait_for_free nf.t.tx_client numneeded` does if you ask for more slots than exist in the ring.

hannesm:
since our different ethernet backends behave differently when a frame bigger than their MTU is passed down, it may be worth unifying this behaviour, and specifying it as documentation in the [module interface](https://github.com/mirage/mirage-net/blob/master/src/mirage_net.mli)?

avsm:
This one didn't make it into the 3.1.1 release, so we'll need to cut a minor 3.1.2 with it soon to unbreak virtio.

djs55:
I'll merge this as-is and release a quick v3.1.2 to give us time to fix this properly.

djs55:
@avsm @hannesm https://github.com/ocaml/opam-repository/pull/9172

PR #313, opened 2017-05-16
Test mtus

yomimono:
The first commit adds tests for the interaction between the TCP implementation 
and the MTU reported by the IP module over which TCP is parameterized.
The following tests (along with some necessary infrastructure and refactoring) are added:

* stacks with different MTUs can connect via TCP
* stacks with different MTUs can make it through the iperf tests
* calls to TCP.write with Cstruct.t's > MTU don't result in attempted
  sends of Ethernet frames larger than the MTU
* calls to TCP.write with large Cstruct.t's don't violate the other
  side's signaled MSS
* calls to TCP.write with large Cstruct.t's don't send multiple small
  Ethernet frames when both sides have large MTUs and MSSs

A subset of tests in the first commit fail on tcpip 3.1.0 and 3.1.1 (excessively large packets can be sent).  A subset of tests in the first commit fail on tcpip 3.1.2 (excessively small packets are sent when both sides could send larger ones).

The second commit causes all tests to pass.

The tests require the third commit, a pin of a branch of [mirage-vnetif](https://github.com/yomimono/mirage-vnetif/tree/mtu-enforcing-netifs) which fails when `write` is called on a virtual network interface with data larger than the interface can write.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ script: bash -ex .travis-ci.sh
 sudo: required
 env:
   global:
-  - EXTRA_REMOTES="https://github.com/djs55/mirage-dev.git#fix-revdeps"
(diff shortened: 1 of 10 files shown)


djs55:
This looks good to me!

PR #316, opened 2017-06-05
Don't link clients against PPX libraries

djs55:
The package cstruct.ppx contains helper functions for the PPX rewriter itself. Linking this code into applications makes them larger and also brings in a dependency on compiler-libs which pollutes the global module namespace with names like Types.

See https://github.com/mirage/mirage-www/pull/556


yomimono:
@djs55, are you checking for this across the package universe?  If not, do you think it's likely to be widespread enough to merit doing so?  I'm asking because I have bits of shell scripts from the MirageOS 3 release which might be able to save some human time.

djs55:
@yomimono I bet this is quite widespread -- IIRC not many oasis-using packages use the `XMETARequires` (or at least the ones I regularly work on don't). I'll do a bit of looking around now that I (think I) have a handle on the problem. If you have scripts, that would be very handy! :)

yomimono:
I think I just found this in a non-OASIS package too: https://github.com/haesbaert/rawlink/compare/master...yomimono:cstruct-ppx-tags?expand=1

djs55:
... and a mistake I made very recently, oops: https://github.com/mirage/shared-memory-ring/pull/33. On the plus side when I pin `shared-memory-ring`, `tcpip` and `mirage-profile`, the `mirage-www` Xen unikernel links again (*cough*)

yomimono:
The tests are still failing with a problem building the stats page for mirage-www, but I'll post an issue over on the mirage-www repository for that rather than belaboring this PR.  Thanks, @djs55!

PR #319, opened 2017-06-16
port to jbuilder

avsm:
This significantly rearranges the source tree, but preserves
the old META format as much as possible.  It adds a new `tcpip.unix`
which is needed for the unix checksum stubs to be linked in.
Previously those were part of the `tcpip` library, which risks
a double link in the case of the xen and solo5 variants.
    
Build time seems to be improved by 4x with this change...


Makefile
```diff
@@ -1,54 +1,16 @@
-# OASIS_START
-# DO NOT EDIT (digest: a3c674b4239234cbbe53afe090018954)
+.PHONY: build clean test
 
-SETUP = ocaml setup.ml
+build:
+	jbuilder build @install
```
samoht: `jbuilder build --dev` is better here

Makefile
```diff
@@ -1,54 +1,16 @@
...
 
-build: setup.data
-	$(SETUP) -build $(BUILDFLAGS)
+test:
+	jbuilder runtest
```
samoht: `jbuilder runtest --dev` is better here

Makefile
```diff
@@ -1,54 +1,16 @@
...
 
-doc: setup.data build
-	$(SETUP) -doc $(DOCFLAGS)
+install:
```
samoht: do you use `install` and `uninstall` not sure they are really needed.

Makefile
```diff
@@ -1,54 +1,16 @@
...
-
-release:
-	git tag -a v$(VERSION) -m "Version $(VERSION)."
-	git push upstream v$(VERSION)
-	$(MAKE) pr
-
-pr:
-	opam publish prepare $(NAME).$(VERSION) $(ARCHIVE)
-	OPAMYES=1 opam publish submit $(NAME).$(VERSION) && rm -rf $(NAME).$(VERSION)
+	rm -rf _build *.install
```
samoht: `jbuilder clean` is better here.

Also it could make sense to add the magic runes for multi-package releases that we use elsewhere.

src/arpv4/jbuild
```diff
@@ -0,0 +1,6 @@
+(library
```
samoht: missing `(jbuild_version 1)` header

src/ethif/jbuild
```diff
@@ -0,0 +1,7 @@
+(library
```
samoht: missing `(jbuild_version 1)` header

src/tcpip_checksum/jbuild
```diff
@@ -0,0 +1,30 @@
...
+  (libraries   (cstruct))
+  (wrapped false)))
+
+(library
+ ((name tcpip_xen)
+  (public_name tcpip.xen)
+  (libraries (tcpip))
+  (modules (tcpip_xen))
+  (c_names (checksum_stubs_xen))
+  (c_flags (:include c_flags_xen.sexp))
```
samoht: does this already work with a released version of `jbuilder`?

Makefile
avsm: its not --dev clean yet, due to various warnings. will add once its clean on all supported versions

Makefile
avsm: i dont; @djs55 i think this came from you... do you?

Makefile
avsm: done in 93118b054d2ce6245970376a5acf18dd2ac91af7

samoht:
I've added a few comments -- please add `(jbuild_version 1)` everywhere (I am too lazy to comment on every file :p)

I like the 4x speedup!!

src/tcpip_checksum/jbuild
avsm: yep...

Makefile
djs55: I think I use them occasionally. Probably so rarely it doesn't matter though.

src/tcpip_checksum/jbuild
samoht: ha yes https://github.com/janestreet/jbuilder/pull/153 fixes it for `flags` but it was already working for `c_flags`. Great :-)

src/ethif/jbuild
avsm: done in 7a3b0ea00469878eda8f15b962f9aa5bf24e0770

src/tcpip_checksum/jbuild
djs55: Copied from the quick start: http://jbuilder.readthedocs.io/en/latest/quick-start.html?highlight=c_flags.sexp#defining-a-library-with-c-stubs-using-pkg-config

FWIW I think this works with `io-page-xen` -- I put an `#error "woo"` in some code guarded by `#ifdef MINIOS` and it triggered as expected

src/arpv4/jbuild
avsm: done in 7a3b0ea00469878eda8f15b962f9aa5bf24e0770

test/test_iperf.ml
```diff
@@ -194,42 +194,42 @@ let test_tcp_iperf_two_stacks_basic amt timeout () =
...
-    (Printf.sprintf "tests/pcap/tcp_iperf_two_stacks_basic_%d.pcap" amt)
+    (Printf.sprintf "tcp_iperf_two_stacks_basic_%d.pcap" amt)
     (Test.tcp_iperf ~server ~client amt timeout)
 
 let test_tcp_iperf_two_stacks_mtu amt timeout () =
   let module Test = Test_iperf (Vnetif_backends.Mtu_enforced) in
   Test.default_network () >>= fun { backend; Test.client; Test.server } ->
   Test.V.record_pcap backend
-    (Printf.sprintf "tests/pcap/tcp_iperf_two_stacks_mtu_%d.pcap" amt)
+    (Printf.sprintf "tcp_iperf_two_stacks_mtu_%d.pcap" amt)
```
MagnusS: I think this will enable pcap recording by default in CI as the dir will now exist. If CI becomes slow it may be because of the large files (unless the pcap recording has been fixed to record less data since then)

yomimono: pcap recording still grabs the full trace, as far as I know.  We should probably rig up a proper command-line argument for recording these traces (and allow the user to specify a directory there too).  @samoht how hard would that be to do with alcotest?

samoht: @yomimono it is not possible to add exta CLI arguments to alcotest yet, but it should be doable without much trouble. Would you mind opening an issue in the alcotest repo to keep track of that? I might add this shortly as I will need to make a new jbuilder release anyway.

yomimono: Sure, https://github.com/mirage/alcotest/issues/93 .  Thanks @samoht.

PR #325, opened 2017-08-01
Improve Tcp.Wire module

samoht:
- add doc strings
- rename `type id` into `type t`
- rename the `wire` creator function into `v`
- expose `src` to get the local IP


yomimono:
Thanks for this!

Issue #324, opened 2017-07-06
linking against ipv4 module fails with missing checksum symbol


yomimono:
This is visible in the Travis output for #320 , but I've reproduced it locally as well:

```
home/dothraki/.opam/4.05.0+trunk+afl/lib/tcpip/ipv4/tcpip_ipv4.a(ipv4_packet.o): In function `camlIpv4_packet__unsafe_fill_1734':
/home/dothraki/.opam/4.05.0+trunk+afl/build/tcpip.3.1.4/_build/default/src/ipv4/ipv4_packet.ml:59: undefined reference to `caml_tcpip_ones_complement_checksum'
/home/dothraki/.opam/4.05.0+trunk+afl/lib/tcpip/ipv4/tcpip_ipv4.a(ipv4_packet.o): In function `camlIpv4_packet__check_1879':
/home/dothraki/.opam/4.05.0+trunk+afl/build/tcpip.3.1.4/_build/default/src/ipv4/ipv4_packet.ml:145: undefined reference to `caml_tcpip_ones_complement_checksum_list'
/home/dothraki/.opam/4.05.0+trunk+afl/lib/tcpip/tcpip.a(tcpip_checksum.o): In function `camlTcpip_checksum__fun_1277':
:(.text+0x37): undefined reference to `caml_tcpip_ones_complement_checksum_list'
/home/dothraki/.opam/4.05.0+trunk+afl/lib/tcpip/tcpip.a(tcpip_checksum.o): In function `camlTcpip_checksum__fun_1279':
:(.text+0x87): undefined reference to `caml_tcpip_ones_complement_checksum'
/home/dothraki/.opam/4.05.0+trunk+afl/lib/tcpip/tcpip.a(tcpip_checksum.o): In function `camlTcpip_checksum__2':
:(.data+0x58): undefined reference to `caml_tcpip_ones_complement_checksum_list'
:(.data+0x60): undefined reference to `caml_tcpip_ones_complement_checksum'
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
Command exited with code 2.
pkg.ml: [ERROR] cmd ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-tag' 'debug'
     '-build-dir' '_build' 'pkg/META' 'CHANGES.md' 'LICENSE.md' 'README.md'
     'src/fuzz_ethif.native' 'src/fuzz_ipv4.native' 'src/fuzz_icmpv4.native'
[...]

PR #332, opened 2017-08-03
Almost but not quite fix Xen linking problem

djs55:
This patch adds a `xen_link_opts` to the META file for `tcpip` which causes
the `mirage` tool to add

'-L/home/opam/.opam/4.04.2/lib/tcpip/xen' '-ltcpip_xen_stubs'

to any unikernel which uses the `tcpip.xen` package. A separate patch will
be needed to link `tcpip.xen` for Xen targets.

Part of #324

### DIFF
--- a/META.tcpip.template
+++ b/META.tcpip.template
@@ -1,0 +1,3 @@
+# JBUILDER_GEN
+
+xen_linkopts = "-L@tcpip/xen -ltcpip_xen_stubs"


samoht:
This should be picked-up by the recursive query. Do you have a setup that you could push to docker hub? Or I will try to quickly setup something on my mac.

djs55:
I'll give you a `Dockerfile` in a moment. My description was wrong -- it does provide `-Ltcpip` but not `-Ltcpip/xen` (where the `xen` stubs are)

samoht:
In that case I think you can just add `-L@name/xen` to the ocamlfind flag. The mirage tool will expand `@name` to the right library name (hopefully)

djs55:
thanks!

yomimono:
Thanks @djs55, this works with the `mirage` changes I mentioned in #324.  Excelsior!

Issue #331, opened 2017-08-01
socket test is failing on windows


samoht:
```
$  _build/default/test/test.exe test socket 1
[..]
test.exe: [INFO] Manager: connect
test.exe: [INFO] Manager: configuring
test.exe: [WARNING] Manager: sockets currently bind to all available IPs. IPs 127.0.0.1 were specified, but this will be ignored
test.exe: [INFO] Manager: connect
test.exe: [INFO] Manager: configuring
test.exe: [WARNING] Manager: sockets currently bind to all available IPs. IPs 127.0.0.1 were specified, but this will be ignored
[exception] Unix.Unix_error(Unix.EPROTONOSUPPORT, "socket", "")
```

yomimono:
This was fixed in #333 AFAIK, thanks for the report!

PR #333, opened 2017-08-04
Fix the ICMP socket test on Windows

djs55:
This patch works around some Windows-isms which currently leak through the lower-level libraries:

- Windows uses `SOCK_RAW` for ICMP sockets whereas Unix uses `SOCK_DGRAM`
- `Lwt` on Windows doesn't implement `Lwt_bytes.{recvfrom,sendto}`

At module load time we compute `Sys.os_type = "Win32"` and then use this to select between Unix and Windows.

Ideally these minor differences would be abstracted away by the lower-level libraries.

Fixes the specific case mentioned in #331 although there is still another failing test.

### DIFF
--- a/src/stack-unix/icmpv4_socket.ml
+++ b/src/stack-unix/icmpv4_socket.ml
@@ -31,6 +31,24 @@ let pp_sockaddr fmt sa =
 let src = Logs.Src.create "icmpv4_socket" ~doc:"Mirage ICMPv4 (Sockets Edition)"
 module Log = (val Logs.src_log src : Logs.LOG)
 
+let sendto' fd buf flags dst =
+  if is_win32 then begin
+     (* Lwt on Win32 doesn't support Lwt_bytes.sendto *)
+     let bytes = Bytes.make (Cstruct.len buf) '\000' in
+     Cstruct.blit_to_bytes buf 0 bytes 0 (Cstruct.len buf);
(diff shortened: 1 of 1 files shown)


yomimono:
This looks good to me.  Out of curiosity, does Windows allow unprivileged ICMP sockets?

djs55:
Dunno -- I didn't look very closely. It would be interesting to find out though... I'll have a play when I'm next in front of a Windows machine.

djs55:
Thinking about it, when I ran this test it was as a regular user I think. Interesting!

djs55:
I've made a simple `ping` CLI tool (based on the code in the ICMP test case) and it runs as an unprivileged user on both Mac and Windows (on Linux I always get `EACCES` even as root -- not sure what's going on there).

Windows just worked (to my surprise!) but the Mac kernel seems to return an IP datagram with an over-large length (16384). I guess this reflects some kind of maximum theoretical size / kernel buffer size rather than the actual on-the-wire size. Seems a bit buggy; I'll have to work around it somewhere.

I was also surprised that a whole IP datagram was returned, rather than only the ICMP portion. This is useful though for things like the `ttl` so that's good I guess...

I'll make a PR with the example soon.

yomimono:
It's probably worth documenting that you get the whole packet, that's surprising (and I'm not sure intended?)

PR #334, opened 2017-08-04
Possible fix for IPv6 unit test

### LINKED ISSUE #330: ipv6 test is failing on windows
```
$ _build/default/test/test.exe test ipv6 0
[...]
Send a UDP packet from one IPV6 stack and check it is received by another.
_build/_tests\ipv6.000.output:
test.exe: [INFO] Connected Ethernet interface 02:50:00:00:00:01
test.exe: [INFO] IP6: Starting
test.exe: [DEBUG] ND6: Sending RS
test.exe: [DEBUG] ND6: Sending NS src=:: dst=ff02::1:ff00:1 tgt=fe80::50:ff:fe00:1
test.exe: [DEBUG] no size limit, sending
test.exe: [DEBUG] no size limit, sending
test.exe: [DEBUG] ND6: Sending NS src=:: dst=ff02::1:ff00:23 tgt=fc00::23
test.exe: [DEBUG] no size limit, sending
test.exe: [INFO] UDP interface connected on
test.exe: [INFO] Connected Ethernet interface 02:50:00:00:00:02
test.exe: [INFO] IP6: Starting
test.exe: [DEBUG] ND6: Sending RS
test.exe: [DEBUG] ND6: Sending NS src=:: dst=ff02::1:ff00:2 tgt=fe80::50:ff:fe00:2
test.exe: [DEBUG] no size limit, sending
test.exe: [DEBUG] no size limit, sending
test.exe: [DEBUG] ND6: Sending NS src=:: dst=ff02::1:ff00:45 tgt=fc00::45
test.exe: [DEBUG] no size limit, sending
test.exe: [INFO] UDP interface connected on
test.exe: [DEBUG] IP6: Queueing packet: dst=fc00::45
test.exe: [DEBUG] ND6: Sending NS src=:: dst=ff02::1:ff00:45 tgt=fc00::45
test.exe: [DEBUG] no size limit, sending
test.exe: [DEBUG] ND: Received NS: src=:: dst=ff02::1:ff00:45 tgt=fc00::45
test.exe: [DEBUG] SLAAC: fe80::50:ff:fe00:1 --> PREFERRED
test.exe: [DEBUG] SLAAC: fc00::23 --> PREFERRED
test.exe: [DEBUG] SLAAC: fe80::50:ff:fe00:2 --> PREFERRED
test.exe: [DEBUG] SLAAC: fc00::45 --> PREFERRED
------------------------------------------------------------------------------------------------------------------------------
ASSERT UDP packet should have been received
------------------------------------------------------------------------------------------------------------------------------
Test error: Error UDP packet should have been received.
```

djs55:
I am not familiar with either IPv6 or this code, but I think the following could happen:
- client sends an IPv6 packet while the stack is initialising
- the packet would be queued
- an entry in the `IpMap` would be created mapping to `INCOMPLETE`

This means that:
- all future writes would call `query`
- which would call `IpMap.find`
- which would *not* raise `Not_found`
- which means that we wouldn't send an "NS" message
- which means that no traffic would ever be sent to that destination

If the initial `sleep` in the unit tests was too short, then the test would timeout. Windows has different timing and the test still failed even with a 1s `sleep`.

This PR:
- removes the `sleep` from the IPv6 UDP unit test
- instead, makes the sender send a packet every 50ms and the test succeeds when the first packet arrives
- *experimental* *danger* attempts to fix the underlying bug by sending "NS" messages (whatever they are) when an IP address is mapped to `INCOMPLETE`

Fixes #330


djs55:
OK I think the relevant part of [RFC2461](https://tools.ietf.org/html/rfc2461) is 7.2.2:

>   While awaiting a response, the sender SHOULD retransmit Neighbor
>   Solicitation messages approximately every RetransTimer milliseconds,
>   even in the absence of additional traffic to the neighbor.
>   Retransmissions MUST be rate-limited to at most one solicitation per
>   neighbor every RetransTimer milliseconds.
>
>   If no Neighbor Advertisement is received after MAX_MULTICAST_SOLICIT
>   solicitations, address resolution has failed.  The sender MUST return
>   ICMP destination unreachable indications with code 3 (Address
>   Unreachable) for each packet queued awaiting address resolution.

It looks like this is supposed to happen via the [tick](https://github.com/mirage/mirage-tcpip/blob/7505d203fa6f40a0ce12402a46ccf844e3bb3f0e/src/ipv6/ndpv6.ml#L521) function. I'll investigate to see why this isn't working.

djs55:
OK, I now believe I've fixed this properly. The problem seems to be that the initial timeout when a `NeighborCache` entry is set to `INCOMPLETE` is erroneously based on the "Reachable Time" which is the time to assume a confirmed table entry is valid instead of the "Retrans timer" which is the expected interval between initial transmitted probes.

The second and subsequent probes would have their timeouts set correctly to `now + retrans_timer` but the initial probe has a bad timeout of `now + reachable_time`. Since `reachable_time` is of the order of 20-30s it causes excessive packet loss and breaks the test. Using the expected 1s `retrans_timer` fixes the test.

yomimono:
I won't have time today to page enough ipv6 stuff into my head to review this properly, but I'll try to get to it soon.

djs55:
No worries -- I don't think it's particularly urgent. I mainly investigated to learn a little about IPv6, to prepare myself for the day when my broadband provider finally rolls it out (any year now!)

hannesm:
what is the status of this PR? does it improve the current IPv6 implementation? if yes, I'd be in favour of merging (esp. since I've some API changes and need to adapt ipv6.ml as well and am scared of merge conflicts). :)

djs55:
I'm fairly confident it fixes a real bug in the IPv6 implementation but I'm much less familiar with it than I am with the IPv4 implementation. However I think we should probably merge this (and figure out how to use and test IPv6 more in future)

PR #335, opened 2017-08-05
Add a dependency from tcpip to tcpip.unix

djs55:
This is obviously a hack and conceptually wrong but

- it fixes builds with the current mirage tool
- it is the same behaviour as before the recent package restructuring

This works because

- the mirage tool uses `ocamlopt` to link Unix binaries, which will include the "Unix" stubs
- the mirage tool manually links other backends, and ignores the stubs referenced by the OCaml object files

Whenever we have proper cross-compilation for mirage backends we can merge together the `tcpip` `tcpip.unix` and `tcpip.xen` packages.

Related to [mirage/mirage#856]

### DIFF
--- a/src/tcpip_checksum/jbuild
+++ b/src/tcpip_checksum/jbuild
@@ -4,7 +4,7 @@
  ((name        tcpip)
   (public_name tcpip)
   (modules (tcpip_checksum))
-  (libraries   (cstruct))
+  (libraries   (cstruct tcpip.unix))
   (wrapped false)))
 
 (library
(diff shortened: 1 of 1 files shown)


djs55:
I successfully built all of [mirage-skeleton](https://github.com/yomimono/mirage-skeleton/tree/name-changes) with `unix`, `xen`, `ukvm` and `virtio` backends with the following pins:
```
$ opam pin
charrua-core.0.8          git  git://github.com/mirage/charrua-core          
mirage.3.0.4              git  git://github.com/yomimono/mirage#name-changes 
mirage-runtime.3.0.0      git  git://github.com/yomimono/mirage#name-changes 
mirage-types.3.0.0        git  git://github.com/yomimono/mirage#name-changes 
mirage-types-lwt.3.0.0    git  git://github.com/yomimono/mirage#name-changes 
tcpip.3.2.0               git  git://github.com/djs55/mirage-tcpip#unix-stubs
```
so I think this is good.

PR #336, opened 2017-08-08
Add user-space ping example

djs55:
This PR includes a simple `ping` implementation which uses the user-space socket stack based on `test/test_icmpv4.ml`:
```
$ ./_build/default/examples/ping/ping.exe -c 5 8.8.8.8
56 bytes from 8.8.8.8: icmp_seq=0 ttl=58 time=10.698080 ms
56 bytes from 8.8.8.8: icmp_seq=1 ttl=58 time=9.099007 ms
56 bytes from 8.8.8.8: icmp_seq=2 ttl=58 time=10.119915 ms
56 bytes from 8.8.8.8: icmp_seq=3 ttl=58 time=10.429859 ms
56 bytes from 8.8.8.8: icmp_seq=4 ttl=58 time=9.947062 ms
--- 8.8.8.8 ping statistics ---
5 packets transmitted, 5 packets received, 0% packet loss
round-trip min/avg/max/stddev = 9.099/10.059/10.698/0.609 ms
```
This PR includes some small fixes to the user-space sockets implementation:
- the ICMPv4 `listen` now blocks forever, like other `listen` implementations. Previously it would block until a single packet is received.
- on the Mac the IP datagram returned by the kernel has a bogus total length field which is then rejected by our IPv4 parser. Rather than relax the generic IPv4 parser this is corrected for in the user-space socket stack before calling the `listen` callback.

### DIFF
--- a/examples/ping/jbuild
+++ b/examples/ping/jbuild
@@ -1,0 +1,4 @@
+(executables
+ ((names (ping))
+  (libraries   (cmdliner logs logs.fmt tcpip.icmpv4-socket))
+))
--- a/examples/ping/ping.ml
+++ b/examples/ping/ping.ml
@@ -1,0 +1,178 @@
+
+let src =
+  let src = Logs.Src.create "ping" ~doc:"Mirage ping" in
+  Logs.Src.set_level src (Some Logs.Info);
+  src
+
+module Log = (val Logs.src_log src : Logs.LOG)
+
+(* Construct a payload buffer of a given size *)
+let make_payload ~size () =
+  let buf = Cstruct.create size in
+  let pattern = "plz reply i'm so lonely" in
+  for i = 0 to Cstruct.len buf - 1 do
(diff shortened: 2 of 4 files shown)


yomimono:
Happily you don't get the binary if you `jbuilder build @install`.  I confess to being a little baffled by the current build system here, but I think this is because it's not included in `META.tcpip.template` -- is that correct, @djs55 ?

djs55:
I couldn't remember so I had a quick look at mirage/ocaml-9p which has a `protocol-9p-tool` package and the `jbuild` file has a [stanza](https://github.com/mirage/ocaml-9p/blob/7ef32ccc0bfadb773bd761f48e6018ff8985c2a8/src/jbuild#L8)
```
(install
 ((section bin)
  (package protocol-9p-tool)
  (files ((main.exe as 9p-tool)))))
```
So I think without one of these `install` sections the binary won't be included.

PR #338, opened 2017-09-05
Add support for TCP keepalives, for both socket and direct

djs55:
This is an implementation of https://github.com/mirage/mirage-protocols/pull/8

The socket implementation uses `setsockopt` on Unix-like systems and `WSAIoctl` on Win32.

The direct implementation uses the same algorithm, implemented in the `keepalive` module. There are tests for the core logic (which is free of I/O) and an end-to-end test of the TCP stack with a broken network.

This PR also includes
- a rename of `Sequence.incr` to `Sequence.succ` to follow the pattern used by `Int32`, `Int64` etc
- merging the TCP `Pcb` and `Flow` modules since `Flow` was only a thin wrapper over `Pcb`


djs55:
Looks like a `-safe-string` problem:
```
# File "test/test_deadlock.ml", line 114, characters 42-46:
# Error: This expression has type string but an expression was expected of type
#          bytes
```

PR #341, opened 2017-11-05
Fix build with OCaml 4.06 (and with -safe-string)

djs55:
The names of some `Ipaddr` and `Macaddr` functions are misleading;
when they say "bytes" they mean the encoding within the value, not
the type of the value and not the mutable/immutable-ness.

- `Ipaddr.V4.to_bytes`: this returns a `string` containing a byte
  encoding of the address (not a `bytes` containing a byte encoding
  of the address)
- `Macaddr.of_bytes_exn`: this consumes a `string` containing a
  byte encoding of the MAC address (not a `bytes` containing a
  byte encoding)
- `Macaddr.to_bytes`: this returns a `string` containing a byte
  encoding of the MAC address


yomimono:
Thanks @djs55 !

PR #343, opened 2017-11-12
Fix tcp deadlock

mfp:


yomimono:
This seems good to me but I'd appreciate a second opinion; @hannesm ?

yomimono:
I'm deploying this on a testing basis and will merge if it seems OK.

djs55:
@yomimono did you have any luck with your deployment?

yomimono:
Yes, it seems OK, so merged.  Thanks for the nudge!

PR #350, opened 2018-02-14
Fix build with OCaml 4.06 (and with -safe-string) again

djs55:
I think this slipped through because the PR build didn't have 4.06 in
the travis matrix, but master does.

Signed-off-by: David Scott <dave.scott@docker.com>


djs55:
```
===== ∗  10   ↻  14 =====
Fatal error: exception File "src/ci_opam.ml", line 116, characters 36-42: Assertion failed
```

PR #353, opened 2018-02-21
Add missing deps

rgrinberg:
Before beta18 dune was lax about this because it built all the objects in a single directory. Beta 18 is no longer so permissive.


djs55:
I restarted the 2 failed travis jobs since they'd failed for spurious reasons AFAICT.

Issue #355, opened 2018-03-02
linking against the socket stack is broken in 3.4.0


yomimono:
See https://github.com/mirage/ocaml-git/pull/280#issuecomment-370056975 for the original report (thanks @dinosaure ).

After `mirage configure -t unix --net=socket` and `make depend`:

```
~/mirage-skeleton/device-usage/network$ make
mirage build
ocamlfind: Package `tcp_socket_options' not found - required by `tcpip.tcpv4-socket'
ocamlfind ocamldep -package tcpip.unix -package tcpip.udpv4-socket -package tcpip.tcpv4-socket -package tcpip.stack-socket -package tcpip -package mirage-unix -package mirage-types-lwt -package mirage-types -package mirage-runtime -package mirage-logs -package mirage-clock-unix -package lwt -package functoria-runtime -predicates mirage_unix -modules main.ml > main.ml.depends
+ ocamlfind ocamldep -package tcpip.unix -package tcpip.udpv4-socket -package tcpip.tcpv4-socket -package tcpip.stack-socket -package tcpip -package mirage-unix -package mirage-types-lwt -package mirage-types -package mirage-runtime -package mirage-logs -package mirage-clock-unix -package lwt -package functoria-runtime -predicates mirage_unix -modules main.ml > main.ml.depends
ocamlfind: Package `tcp_socket_options' not found - required by `tcpip.tcpv4-socket'
```

PR #356, opened 2018-03-02
fix socket build; test unix target with --net=socket as well

yomimono:
Without this, we have no tests making sure a user can link against the socket stack, which makes it hard to realize that we've broken linking against the socket stack.

Incidentally, [we've broken linking against the socket stack](https://github.com/mirage/mirage-tcpip/issues/355).


yomimono:
@rgrinberg I'm not sure whether the fix I've included here works by design or by accident - can you just drop a quick LGTM if it doesn't look too broken to you, or a pointer at some relevant doc or example to guide me in fixing it if it's wrong?

rgrinberg:
Hi Mindy, the fix you've included only works because you've added a public_name to this options library. You can keep the old names when specifying dependencies, dune doesn't care which ones you use.

Also, changing the order of defined libraries shouldn't make a difference. If it does, please file a bug.

yomimono:
> Also, changing the order of defined libraries shouldn't make a difference. If it does, please file a bug.

Cool.  It's easier for me to make sense of a complicated build like this one if all the names refer to something I've already seen, so I'm keeping this reordering.

> You can keep the old names when specifying dependencies, dune doesn't care which ones you use.

Thanks, I've updated the change to remove the extaneous renaming.

Do I infer correctly from the comments in https://github.com/ocaml/dune/pull/580 that all of the releases of `tcpip` using jbuilder (3.2.0, 3.3.0, 3.3.1, and 3.4.0) will have this problem, and no version constraint on `jbuilder` will fix it?

rgrinberg:
Yes. This was always a limitation of dune. Note that the next version of
dune will not even lift the limitation but will have a better error message
for it. However, we'll lift this limitation eventually

hannesm:
thanks for fixing this!

PR #364, opened 2018-06-02
Tcp_socket: avoid raising an exception (`Lwt.fail exn`) when `write` …

hannesm:
…errors,

  instead use the variant ``[ `Exn of exn ]`` in (`Lwt.return (Error (`Exn exn))`)
  as done in `read`
Tcpv4_socket / Tcpv6_socket: expose the custom variants in `type error` and
  `type write_error` in the interface

last time the semantics changed was in 47c8827eb9a488578c4cd41e3fbeb0451ec29bf8 by @djs55 -- I believe since Mirage 3.0 is now out of the door, where we claim to unify error handling by using the result type (i.e. `(data, err) result Lwt.t` instead of `data Lwt.t`), we should adapt the `write` on unix as well.

I don't know whether `` [ `Exn of exn ] `` is a nice type to work with (`Mirage_protocols.Tcp.error` contains ``[ `Timeout | `Refused ]``, whereas the `write_error` adds `` [ `Closed ]`` -- not sure how to best map all the unix socket errors into that sum type (it may need some refinement at a later point).

The purpose of this PR is to unify the MirageOS-as-unikernel-on-a-hypervisor with the Unix backend.  I hate to have to add random noise of `Lwt.catch` in my code after I discover that on Unix I can still get `Lwt.fail` instances.

[EDIT: removed the `close` question, opening another issue for discussing this instead.]

### DIFF
--- a/src/stack-unix/tcp_socket.ml
+++ b/src/stack-unix/tcp_socket.ml
@@ -2,13 +2,15 @@ open Lwt
 open Result
 
 type error = [ Mirage_protocols.Tcp.error | `Exn of exn ]
-type write_error = Mirage_protocols.Tcp.write_error
+type write_error = [ Mirage_protocols.Tcp.write_error | `Exn of exn ]
 
 let pp_error ppf = function
   | #Mirage_protocols.Tcp.error as e -> Mirage_protocols.Tcp.pp_error ppf e
-  | `Exn e -> Fmt.pf ppf "%a" Fmt.exn e
+  | `Exn e -> Fmt.exn ppf e
 
-let pp_write_error = Mirage_protocols.Tcp.pp_write_error
+let pp_write_error ppf = function
+  | #Mirage_protocols.Tcp.write_error as e -> Mirage_protocols.Tcp.pp_write_error ppf e
+  | `Exn e -> Fmt.exn ppf e
 
 let disconnect _ =
   return_unit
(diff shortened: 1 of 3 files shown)


samoht:
Do you know which kind of errors can be raised by `write`? I am not against catching the errors but I don't see the difference between `Lwt.raise` and `Lwt.return (`Exn exn)`.

hannesm:
@samoht errors which can be raised by `write` (which calls `Lwt_cstruct.write`, which calls `Lwt_bytes.write`, which calls POSIX `write`) are documented at http://pubs.opengroup.org/onlinepubs/009695399/functions/write.html
The issue I observed was when I had a TCP connection where the other end died (without terminating the connection), and I got a `CONNRESET` (via `Lwt.fail`).  I only ran the code with the direct stack before, where `write` never throws `Lwt.fail`, but uses a `result` type (as documented in `FLOW` at https://docs.mirage.io/mirage-flow/Mirage_flow/module-type-ABSTRACT/index.html#val-write).

Now, the suggested PR adjust the socket stack to meet the `FLOW` specification -- instead of `Lwt.fail` for an error, it uses `Lwt.return (Error e)` --> code which is programmed against the `FLOW` API does not need to guard `FLOW.write` with `Lwt.catch`.

samoht:
Thanks, this sounds convincing :-)

Issue #367, opened 2018-06-15
Excessive memory consumption when keepalives are enabled


djs55:
The keepalives implementation in #338 maintains a reference to the TCP connection state while the keep-alive timer is set. When the connection is closed the [refresh](https://github.com/mirage/mirage-tcpip/blob/f58e93b3d4ecd1d10530c0b9aedd717c1cb692e7/src/tcp/keepalive.mli#L56) function will not be called as there is no more traffic, so the keep-alive timer will start counting down and sending keep-alive messages. This is unnecessary because the connection has been closed. The peer will send `RST` packets but we ignore those too. Eventually the final timer expires and the keep-alive code gives up and the connection state can be collected by the GC.

Unfortunately the connection state can be quite large, as it includes buffers of data which have been received but not yet transmitted to the application.

Therefore if

- the stack has keep-alive enabled; and
- connects or accepts large numbers of short-lived connections; and
- lots of data is received by the stack and buffered

then the heap usage of the stack can quickly grow to unacceptable levels, causing the machine to swap or otherwise run out of memory.

This was first reported in [moby/vpnkit#371]

As discussed [here](https://github.com/mirage/mirage-tcpip/pull/366) the plan is to add a warning to the keep alive code in the short term and then fix it in a future release.

PR #368, opened 2018-06-15
Emit a warning if TCP keepalives are used

djs55:
As reported in #367, the implementation of keepalives can use unexpected amounts of memory under some circumstances. Until the fix is ready, emit a single warning if a client enables keep-alives.


hannesm:
thanks for doing that, this looks good to me!

PR #370, opened 2018-07-29
Start preparation for Lwt.async demanding unit Lwt.t

cfcs:
Please see https://github.com/ocsigen/lwt/issues/603

This patch ignores the `('a,'e) result` from `q.xmit`
(which was also unhandled previously, so I don't think I have changed the behavior of this code).

### DIFF
--- a/src/tcp/segment.ml
+++ b/src/tcp/segment.ml
@@ -367,7 +367,9 @@ module Tx (Time:Mirage_time_lwt.S) (Clock:Mirage_clock.MCLOCK) = struct
             let flags=rexmit_seg.flags in
             let options=[] in (* TODO: put the right options *)
             Lwt.async (fun () ->
-                q.xmit ~flags ~wnd ~options ~seq rexmit_seg.data >|= ignore);
+                q.xmit ~flags ~wnd ~options ~seq rexmit_seg.data
+                (* TODO should this return value really be ignored? *)
+                >|= fun (_: ('a,'b) result) -> () );
             Lwt.return_unit
           end else
             Lwt.return_unit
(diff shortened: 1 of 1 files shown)


hannesm:
then you should merge it. I don't understand these intrinsics of the tcp implementation too well, and am a bit worried about ignoring some result value... not sure whom `q.xmit` should report its failure (though, atm, it doesn't report it either, so the proposed change preserves semantics) :)

emillon:
What do you think about adding an explicit type annotation? This will catch the case where we add an argument to the function, or change its return type.

cfcs:
@emillon Good suggestion, I'll update with that!

cfcs:
@avsm Do you remember what this function (`xmit`) actually does?

@hannesm suggested that we could make it log when we get a `Error _`, but I can't figure out if this is supposed to fail, and I felt reluctant to add it in case it was supposed to fail 10k times per second in some cases. If it is not supposed to fail often, I think adding a warning message to the log sounds like a very sensible suggestion, especially if it is supposed to *never* fail and there are bugs to be fixed hiding in here.

But I feel like I have zero insight into how this codebase works, and I really just wanted to pave the way for the new `lwt` interface with this PR, so I am hoping I can find someone who feels a bit more for this repository to chime in with their expectations to this patch.

avsm:
I still need to review this and refresh my memory about the implications...

emillon:
I think that the conservative option is to fix the API breakage and add the `>|=` with the type annotation.
I opened #392 to track the question of whether it should be ignored or not - we can replace the TODO by a link to that issue (this was already ignored before).

cfcs:
So, since this patch doesn't change semantics of the current code, should we merge it now?

Then the decision on whether or not we're handling potential `xmit` failure adequately can be made in #392 ?

ping @avsm

cfcs:
https://github.com/ocsigen/lwt/issues/603#issuecomment-523139684

avsm:
This can indeed be released now; thanks!

PR #372, opened 2018-09-02
remove open Result / Result.result types (since June 2017 tcpip requires ocaml 4.03, which provides result in Pervasives)

hannesm:
and tcpip opam file does not specify the `result` package as dependency...  this should be safe! :)

in addition, upcoming OCaml 4.08.0 will contain a more extensive `Result` module, which may have unwanted effects (shadowing of names) if we open it globally ;)

### DIFF
--- a/src/arpv4/arpv4.ml
+++ b/src/arpv4/arpv4.ml
@@ -16,7 +16,6 @@
  *)
 
 open Lwt.Infix
-open Result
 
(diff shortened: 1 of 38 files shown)


hannesm:
can I get someone to approve this? //cc @mirage/core

avsm:
lgtm!

PR #373, opened 2018-09-02
IPv4 reassembly

hannesm:
I encountered an issue when my local (on my laptop) MirageOS VM talks with my remote MirageOS VM (on my server), that when I'm behind a DSL line (with an MTU of 1492), I get "couldn't decode packet" on the server side (this is a _high level error from my DNS implementation_).  Now, investigation lead to the fact that my router fragments the 1514 bytes big ethernet frame into two, which are then sent to my server.

Since mirage-tcpip lacked any reassembly support (it did not even check "more fragments" or "offset" fields), it pushes the first received (partial!) frame up the stack, which leads to something trying to decode an incomplete frame.

I tried various things to fix this issue:
- first, providing manually an MTU on the client side (via modifying `main.ml` to integrate a `~mtu:1492` to `Ethif.connect`). this worked, but it is a bit tedious to tell my unikernels that they've to use a different MTU now only because the path changed
- then loudly dropping all fragmented frames - this didn't lead to a successful communication
- implemented reassembly in IPv4 as: you have to receive all segments in-order, then reassembly is easy.  on any out-of-order segment, choke (see 5bbef18)
- implemented reassembly following the strategy that there may be overlapping and out-of-order segments, taking always the newest one (after re-reading https://eprint.iacr.org/2015/1020.pdf), see d81c7ac 
- implemented reassembly again with out-of-order but disallow overlaps, as done by linux (since https://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git/commit/?id=c30f1fc041b74ecdb072dd44f858750414b8b19f, their recent FragmentSmack mitigation), see a8cc16c
- (to be clear, I'm not satisfied with the initial two attempts, they should be squashed before merge) 

all these three implementation affect the receiving side only, we never manually fragment any outgoing packet (TCP does this for us, UDP does not at all).  There is always a LRU in use, which is atm limited to 4MB of fragments (see the `weight` function taking only the payloads into account, and the `Static_ipv4.connect` body containing the 1024*1024*4).  There is another limit of maximal 16 fragments for each IP frame (copied from my FreeBSD machine) -- certainly both should be user-configurable at some point.  In addition, our stack should use Path MTU discovery to adjust their MTU, but that is slightly trickier since it needs to cache IP endpoint <-> MTU numbers, and deal with ICMP.

This PR lacks tests, but I'd appreciate any comments and feedback on the reassembly implementation, esp. if you're concerned about attacks (I suspect DoS is the only one applicable here).  I'd also be interested if there are any views how and where such configuration variables (which should be per-unikernel!?) should be kept and how they're gonna be adjusted in a deployment scenario (same is true for the `mtu` parameter to `Ethif.connect`, I'm not sure whether I already know this at build-time).

I added two fields to `Ipv4_packet.t`, `id` and `off` -- and I'm actually curious why/how `id` is supposed to work, atm I can see no write to it ever, so it'll always be uninitialised / 0?!? - EDIT: it is actually written (using the value `Random.int 65535` in `Ipv4_common.adjust_output_header`).


hannesm:
NB: not sure about the hardcoded 16 maximum fragments - IPv4 packets can be up to 2^16 - 1 = 65535, which with a standard MTU of 1460 would lead to 45 fragments - should we allow this?

separate note: the ICMP implementation atm replies with the same body for ICMP echo request, this used to be fine, but now it copies the body of the reassembled packet - which leads to a write call which is much bigger than the MTU.  This behaves badly on solo5-virtio (which asserts that the send packet does not exceed 1526 bytes), and solo5-ukvm (which asserts that there are no partial writes, I get partial writes for > 16000 bytes).  looks like write will need to fragment to solve the above mentioned issues.

avsm:
I did wonder about IP reassembly when doing the original stack, but disregarded it in favour of a simpler implementation since there are many virtual usecases where you never see IP fragmentation (and it's largely gone in IPv6 as well).

So I do think that making this optional would be useful, but not necessarily a blocker to merging it.  Upon first read, is there a timeout missing for the fragment cache? The LRU could fill up with lots of never-timed-out fragments from other connections couldnt it?

hannesm:
@avsm I agree that there's no need for IP reassembly if we'd have path MTU discovery.  Nevertheless, even IPv6 specified IP reassembly ;)

Timeout: why? There is a LRU in use, which sticks to its limit (at most contain 4MB of data), no need to carry around timeout logic in my opinion. If the LRU gets filled more and more, the least recently used pieces will be discarded automatically, so that it will never grow above its limit.

avsm:
See [RFC 4963](https://tools.ietf.org/html/rfc4963) -- the issue is reuse of the 16-bit ip id field, causing fragments to be potentially mismatched. The only solution is a lot of bandaids, and a timeout in the fragment cache helps to mitigate it somewhat.

hannesm:
@avsm thanks for the pointer.  what I'd suggest as solution:
- IPv4 needs to depend on Clock (the monotonic one)
- When an IPv4 fragment is received (so MF=true or offset <> 0), retrieve `now` = Clock.elapsed_ns
- Store the timestamp of the first received fragment in the fragments cache
- When there's no entry for the IPv4 "flow", put `now` in the cache
- When there's already an entry, compute the difference between `now` and the stored timestamp, if it is "too big" (according to RFC4963 this is 15 seconds):
  - discard the stored fragments of the flow
  - store the just received fragment together with `now`
- if the timestamp difference is less than 15 seconds, store the received fragment and continue

this way, we don't need timers iterating over the cache every n seconds, neither timeouts/wakeups of tasks.  the only danger is that the fragment cache grows "too big", but we've a (hardcoded) limit in size, so that should be fine.

does this sound appealing?  I use a similar strategy in my [DNS resolver cache](https://github.com/roburio/udns/blob/8a888ad2d00b387a9003d7dd10ef7ff3cb2cc86b/resolver/dns_resolver_cache.ml#L84-L110) (using a LRU and inserting a timestamp, which is checked when retrieving elements).

PR #374, opened 2018-09-15
use full range for sequence numbers

hannesm:
(on top of #371, relevant commit is only the last one)


src/tcp/flow.ml
```diff
@@ -728,10 +717,8 @@ struct
 
   (* Construct the main TCP thread *)
   let connect ip clock =
-    (* XXX: I've no clue why this is the way it is (10000 + Random
-       ~bound:10000) -- hannes *)
     let localport =
-      10000 + (Randomconv.int (fun x -> Random.generate x) ~bound:10000)
+      1024 + (Randomconv.int ~bound:(0xFFFF - 1024) Random.generate)
```
djs55: This patch looks like a definite improvement to me!

Out of curiosity, do we have to choose local ports > 1024? Is this a Unix convention or a TCP convention? (No big deal, I'm just interested)

hannesm: oh, of course any port number is fine, according to the RFCs[tm].  in practise, there are servers which refuse to service you if your source port range is < 1024 (I experienced this with some authoritative nameservers) - thus I'd (for now at least) avoid to use something < 1024 -- but then our UDP implementation takes any source port -- maybe we should unify these two expressions!?

hannesm:
related to this is https://github.com/hannesm/mirage-tcpip/commit/562e6e4262e2007e1834b837b00947a53b48f1c5 together with https://github.com/hannesm/mirage/commit/39e84bdc5270b3d51596c28c067560117a808984 and https://github.com/hannesm/charrua-core/commit/283da1215b33d664a241a67508060a70250b7815 and https://github.com/hannesm/mirage-qubes/commit/743f640649cf3558690db71d9296133fe4c75bb3 -- which adds RANDOM (and MCLOCK, see #373 -- I want to avoid to have to do the same set of releases again next week just because I didn't integrate MCLOCK yet) to the Static_ipv4 interface.  if someone could tell me that's the right direction or not, I'm happy to incorporate these 4 commits __tomorrow__ in the (anyway) pending releases (towards 3.2.0) /cc @mirage/core (comment here, or use the emojis ;).

the issue at hand is: the IPv4 stack (Static_ipv4, via Ipv4_common) still uses OCaml's Random number generator, while we have this nice Mirage RANDOM device, which can be used either with stdlib-random or fortuna/nocrypto.

hannesm:
included in #371, together with the mentioned IPv4 functor adjustment.

PR #375, opened 2018-09-19
IPv4 fragment reassembly

hannesm:
rebased #373 on top of master, and now using timestamps and a maximum time of 10 seconds between arrival of first segment and last segment.

### DIFF
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -1,0 +1,178 @@
+(*
+ * Copyright (c) 2018 Hannes Mehnert <hannes@mehnert.org>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS l SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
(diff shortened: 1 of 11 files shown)


hannesm:
current state:
- lifetime is 10s max between first and last fragment
- size is 1MB hardcoded
- max 16 fragments for each "flow" (source ip, destrination ip, protocol, ipv4 identifier)
- inserted into sorted list, checks overlaps and holes on reassembly (triggered once a fragment without "more fragments" has been received)

this has some issues:
- anyone can spam (with a constant stream of fragmented packets - needs to fill 1MB in 10s) the fragment cache, leading to resource exhaustion of the cache ("valid" fragments are dropped if they're incoming too slowly)
- insertion into linked list is O(n) (with n is maximal 16)
- ping -s 65535 isn't answered with MTU=1500 (doesn't fit into 16 fragments)

what we could do instead
- maximum storage per source ip
- use a bitmask or tree data structure for the segments (offset is on 8byte boundaries)
- may lead to verification of overlaps at insertion time --> can drop immediately

avsm:
Although the DoS is annoying, its better than the current state of affairs isnt it? Since we just drop all fragments now, which is what would happen if the cache was attacked. The only drawback would be the resource usage of the cache, but that's bounded.

hannesm:
avsm: agreed. i added my comment above in fragments.ml and rebased. will merge once CI is happy.

hannesm:
i want to write some unit tests before merging

hannesm:
added 333 test cases which all pass (haven't checked coverage, but should be good) with a handcrafted `permutation` function. once CI passes, I'll merge.

PR #376, opened 2018-10-09
warning cleanups, handle exception in accept

hannesm:
This primarily solves an issue I had with mirage-skeleton's static-website-tls on Unix: `accept` may raise an exception which leads to the termination of the process -- now the behaviour is consistent with the direct stack that a warning is printed, and the loop is called again.

what should nevertheless be done (in a future PR): listen on a single address only (not always on Ipaddr.V4.any), and handle errors from `bind` (we should make that consistent between unix and direct stack, and return a result from `listen`: return EADDRINUSE/EPORTINUSE instead of (unix atm) Lwt.fail, (direct stack atm) Hashtbl.replace port (the previous listener will never receive any packet, but is still around - isn't it?).

### DIFF
--- a/src/stack-direct/tcpip_stack_direct.ml
+++ b/src/stack-direct/tcpip_stack_direct.ml
@@ -129,7 +129,7 @@ module Make
     let t = { netif; ethif; arpv4; ipv4; icmpv4; tcpv4; udpv4;
               udpv4_listeners; tcpv4_listeners } in
     Log.info (fun f -> f "stack assembled: %a" pp t);
-    Lwt.ignore_result (listen t);
+    Lwt.async (fun () -> listen t);
     Lwt.return t
 
   let disconnect t =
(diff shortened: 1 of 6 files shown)


hannesm:
if someone could review and merge, that would be nice.  I plan for the above mentioned EADDRINUSE issue.in a separate PR.

mor1:
FWIW LGTM but probably best to wait for someone who knows this code better...

dinosaure:
LGTM but same as @mor1, I'm not expert about this code.

hannesm:
since this happened again to one of my unix mirageos unikernels, and I re-reviewed the code to ensure the only semantical change is that `recvfrom` and `accept` are now guarded, I'll merge (and draft a new tcpip release (required for next mirage release) tomorrow (friday 16th) at high noon unless someone objects. //cc @mirage/core

Issue #378, opened 2018-11-22
jsoo support


copy:
I'd like to use mirage-tcpip in a project using JavaScript using js_of_ocaml. The only piece that is missing is an OCaml (or JavaScript) implementation of the tcp checksum calculation.

As a quick-and-dirty fix, I ported the C implementation to OCaml and it seems to be working fine: [patch](https://gist.github.com/copy/62a594a668b89a625340e01cbb345604). It's  missing a non-allocating version for the `_list` variant. It'll also need to be properly integrated with the build system so that the native version will be selected for native builds and the pure version of jsoo builds.

avsm:
This sounds good; please do send in a PR with the JS implementation and I'll fix up the build system to integrate it.

samoht:
Would it make sense to extend and use [mirage/checkseum](https://github.com/mirage/checkseum) for this? /cc @dinosaure

PR #383, opened 2018-12-13
Native implementation of checksum for jsoo

copy:
Implement `ones_complement` and `ones_complement_list` in OCaml for support of `js_of_ocaml`.

As per https://github.com/mirage/mirage-tcpip/issues/378#issuecomment-446861272, leaves the native versions disabled for a later fixup.

### DIFF
--- a/src/tcpip_checksum/tcpip_checksum.ml
+++ b/src/tcpip_checksum/tcpip_checksum.ml
@@ -15,6 +15,59 @@
  *)
 
 (** One's complement checksum, RFC1071 *)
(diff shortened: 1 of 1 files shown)


hannesm:
this is not mergeable as is, since it replaces the calls to the checksum stubs (implemented in C) to pure OCaml code -- which would need some performance investigations (whether it is a good idea or not). OTOH, investigation how to make tcpip workable with jsoo would be welcome - esp. if there's a way in the build system to provide custom JS code for the stubs.

dinosaure:
An other possible solution is to use `checkseum` (which provides an OCaml impl. of CRC and be compiled with `jsoo`), but, as you said, we need to check performances.

PR #384, opened 2018-12-27
[WIP] Dunify with variants

avsm:
Prompted by @copy's pure OCaml checksum stubs, I've begun the port of tcpip to use the new Dune variants feature.  This feature in Dune promotes the so-called "linking hack" (where a cmi is shared among multiple implementations) to a first-class supported entity.

We use virtual libraries to break out the checksum handling into a separate `tcpip-checksum` library that has unix, xen and ocaml variants that can be selected at executable link time.  Since each of the implementations is compiled separately, we also have good CFLAGS discipline.

- [ ] where on earth is the `freestanding` version of the checksum stubs at present?  Needs to be added here.
- [ ] document the renaming of the `tcpip` ocamlfind library to `tcpip-checksum` and the use of virtual libraries.
- [ ] remove the duplication of the tcpip_checksum implementations via a copy_file rule.
- [ ] reduce the number of modules that are compiled with `ppx_cstruct` to the ones that really need it (usually just the `wire` libraries).
- [ ] compile the pure OCaml version with jsoo so we can ship a JS tcpip stack. @copy, did you have any usecase for the JavaScript version that we could include as a testcase to verify that it works?
- [ ] dune 1.7 needs to be released before this can be merged.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+### dev
+
+* Port from Jbuilder to Dune and use virtual libraries to cleanup the stub compilation
+  of checksums. (@avsm).
+
 ### v3.7.0 (2019-02-02)
 
 * Use `Lwt_dllist` instead of `Lwt_sequence`, due to the latter being deprecated
@@ -20,6 +25,8 @@
 
 * Use `Ipaddr.pp` instead of `Ipaddr.pp_hum` due to upstream
   interface changes (#385 @hannesm).
+
+* Use `Ipaddr.pp` instead of `Ipaddr.pp_hum` due to upstream interface changes (@avsm).
 
 ### v3.5.1 (2018-11-16)
 
--- a/META.tcpip.template
+++ b/META.tcpip.template
@@ -1,3 +1,0 @@
-# JBUILDER_GEN
-
-xen_linkopts = "-L@tcpip/xen -ltcpip_xen_stubs"
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,2 @@
-(lang dune 1.0)
+(lang dune 1.7)
 (name tcpip)
--- a/examples/ping/dune
+++ b/examples/ping/dune
@@ -1,3 +1,3 @@
 (executables
  (names ping)
- (libraries cmdliner logs logs.fmt tcpip.icmpv4-socket))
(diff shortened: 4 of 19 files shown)


hannesm:
the solo5 checksum stubs are part of mirage-solo5 https://github.com/mirage/mirage-solo5/blob/master/lib/bindings/checksum_stubs.c

I agree with @samoht https://github.com/mirage/mirage-tcpip/issues/378#issuecomment-447277239 and would appreciate to move the checksum code to checkseum and depend on checkseum. this would remove all c stubs from this library :D
for a smooth upgrade path:
 - add crc to checkseum using some `checkseum_` prefix, release checkseum
 - depend on released checkseum, get rid of stubs, release tcpip
 - get rid of checksum stubs in mirage-solo5 (release that and conflict with earlier tcpip)

dinosaure:
I will be happy to improve `checkseum` and provide something for `mirage-tcpip`. However, `checkseum` is on top of [`optint`](https://github.com/mirage/optint.git) which did not have a strong test suite and use a _hack_ about integer representation of `native integer` in the OCaml world.

One, and the biggest, issue will be to test `checkseum` on a 32-bits arch and on a 64-bits.

samoht:
That'd be a good occasion to fix checkseum :-)

avsm:
I don't think we should switch to checkseum just yet -- the two checksumming methods seem entirely orthogonal to each other, and TCP 1s-complement checksumming isn't used very often elsewhere.

This PR maintains compat and introduces a new tcp-checksum package using Dune variants, so I'd prefer to release this first (once Dune 1.7 is out) and then consider moving the stubs out.

avsm:
This depends on dune 1.7 being released, so deferring it for now in favour of https://github.com/mirage/mirage-tcpip/pull/391. Will rebase later

avsm:
Now that dune 1.7 is out I've pushed a rebase of this PR that shifts to virtual_modules. See also mirage/mirage#969

yomimono:
Any chance for another rebase?

hannesm:
now that dune 1.9.0 is out, can we get a rebased PR? :) /cc @TheLortex

avsm:
Holding off on rebasing this until https://github.com/ocaml/dune/issues/2322 is fixed

hannesm:
thanks for your effort -- closing now that we have a slightly revised (xen) stack, where the checksum stubs are provided by mirage-xen themselves.

Issue #349, opened 2018-02-14
Lwt_sequence is deprecated


djs55:
A manual build (with `--dev`) fails with:
```
File "src/tcp/segment.ml", line 32, characters 8-31:
Error (warning 3): deprecated: module Lwt_sequence
 This module is an implementation detail of Lwt. See
   https://github.com/ocsigen/lwt/issues/361
```
Related to [ocsigen/lwt#361]

dinosaure:
Any plan to fix it? I can take my time to try to fix it (needed to `git-mirage`).

yomimono:
@dinosaure I won't have time to fix this in the next couple of weeks and would welcome a PR addressing it.

dinosaure:
Ok. I will make something when @samoht will confirm to me that is the source of bug 👍 .

bikallem:
@dinosaure may I know how you are planning to fix this issue. https://github.com/diml/lambda-term also needs to be updated to use a different doubly linked list. Will you be using a separate implementation of doubly linked list perhaps or per use Lwt_sequence with a different name?

dinosaure:
According to what I said to @yomimono at the last hackathon, my goal is to provide a proved implementation of a double linked-list. I just discover in a [paper](https://hal.inria.fr/hal-00789533/document) than it's not possible to use Why3 in this case. So the plan is to use CFML/Coq instead. So, of course, this library will be outside `mirage-tcpip` and will be used in `lambda-term` too.

yomimono:
See also https://github.com/mirage/mirage/issues/889 , the higher-level tracking issue for lwt 4.0.0 and mirage.

hannesm:
since #388 this no longer uses lwt_sequence.

PR #387, opened 2019-01-05
don't suppress deprecation warnings (3)

hannesm:
as noticed in https://github.com/mirage/mirage-net-xen/pull/79#discussion_r245473869 - I'm actually in favour of not suppressing warnings to remind ourselves that we need to work on this... Lwt_sequence getting deprecated as tracked in https://github.com/mirage/mirage/issues/889

### DIFF
--- a/src/tcp/segment.ml
+++ b/src/tcp/segment.ml
@@ -14,8 +14,6 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-[@@@ocaml.warning "-3"]
-
 open Lwt.Infix
 
 let src = Logs.Src.create "segment" ~doc:"Mirage TCP Segment module"
--- a/src/tcp/user_buffer.ml
+++ b/src/tcp/user_buffer.ml
@@ -15,8 +15,6 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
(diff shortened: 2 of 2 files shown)


avsm:
The issue is that this breaks the default warnings-as-errors mode of dune (which is otherwise reasonable during development). The issue with the Lwt_sequence deprecation is that there's no clear solution to it yet, so the function is marked as deprecated but can't actually be immediately fixed (without pulling out Lwt_sequence into a separate package). So I'd prefer to keep track of these sorts of warnings in issue (as done in mirage/mirage#889) and suppress the deprecation warning until we're ready to fix it.

emillon:
An in-between solution is to explicitly mark the problematic modules with:

```ocaml
[@@@ocaml.warning "-3"]
module Lwt_sequence = Lwt_sequence (* see issue #xxx *)
[@@@ocaml.warning "+3"]
```

BTW do you know if there's a plan to extract the existing `Lwt_sequence` to opam? A verified version would be nice but it does not seem necessary to wait for it.

avsm:
> BTW do you know if there's a plan to extract the existing Lwt_sequence to opam? A verified version would be nice but it does not seem necessary to wait for it.

I did a quick extraction to https://github.com/mirage/lwt-dllist -- as soon as there is agreement on https://github.com/mirage/mirage-tcpip/issues/349 i can cut a release and we can use that instead of the builtin version.

avsm:
fixed by #388 instead

PR #388, opened 2019-01-14
Use `Lwt_dllist` instead of `Lwt_sequence`

avsm:
This is due to the latter being deprecated upstream in Lwt (ocsigen/lwt#361)


hannesm:
bad Github magic automatically closed this one.. would you mind rebasing on master, please?

PR #390, opened 2019-02-02
remove ethernet and arpv4 implementation from this repository

hannesm:
they are now kept in separate repositories (https://github.com/mirage/arp and https://github.com/mirage/ethernet), see https://github.com/mirage/mirage/pull/965 for further details


avsm:
This lgtm. My only comment on the split out repositories is that this might be a good time to enable `wrapped true` hygiene on them.  I'd like to start deprecating the use of `wrapped false` in new code where practical. We can also do this in a later stage...

test/static_arp.ml
```diff
@@ -46,7 +46,7 @@ module Make(E : Mirage_protocols_lwt.ETHIF)(Clock : Mirage_clock_lwt.MCLOCK) (Ti
   let input t buffer =
     (* disregard responses, but reply to queries *)
     try
-    match Arpv4_wire.get_arp_op buffer with
+    match Cstruct.BE.get_uint16 buffer 6 with
```
avsm: Why drop the Arp getters here?

hannesm: because there are none (see https://github.com/mirage/arp/blob/master/src/arp_packet.mli). arp does not use ppx_cstruct atm.

hannesm:
@avsm I agree that in further PRs we should refine and adjust e.g. the Ethernet API, and as well make tcpip (mostly!) independent of a concrete Ethernet implementation (this needs some API adjustments in mirage-protocols, as done in the vlan branch, which is the ultimate reason to do all this work ;).

avsm:
Sounds good. Do you want to me to merge and do a release of ethernet/arp/tcpip now with these changes? I'm doing some other release work now so can do it today.

test/static_arp.ml
avsm: doesn't need cstruct, but just exposing a manual get function might be nice in a future rev. (so no hardcoded integer in tcpip)

hannesm: would be great if you could open an issue on the arp repository with these enhancements, so we don't forget about that -- another possibility would be to provide random generators of arp frames into the arp package instead of here

hannesm:
merged into master without the travis changes.

test/static_arp.ml
avsm: done in https://github.com/mirage/arp/issues/12

PR #391, opened 2019-02-03
release 3.7: port to dune, switch to topkg

avsm:
The dune upgrade was performed with `dune upgrade` from trunk dune


avsm:
thanks for the quick review!

PR #396, opened 2019-02-25
update changes, remove extra remote from travis

hannesm:


hannesm:
although travis fails, I'll merge and release. the failure is while compiling the examples (a new mirage release is necessary, which requires this new tcpip release).

PR #397, opened 2019-03-08
Add IPv4_packet.Unmarshal.header_of_cstruct

yomimono:
Examining only the IPv4 packet header, without making determination
about the validity of the length field, is useful for applications which
would like to parse IPv4 headers in ICMP error payloads.


hannesm:
thanks, this looks good to me. CI is failing since mirage 3.5.0 doesn't work on 4.04 anymore (mirage-types requires mirage-kv >= 2.0.0 which requires >= 4.05.0)

hannesm:
thanks!

PR #401, opened 2019-04-05
when calculating icmpv4 checksum, don't double-count the payload

yomimono:
`into_cstruct` is likely to be passed a buffer larger than `Icmpv4_wire.sizeof_icmpv4`.  Don't consider any additional bytes in the header part of this checksum calculation, as they ought to be included with `payload` in this call.

Fixes https://github.com/mirage/qubes-mirage-firewall/issues/50 .


hannesm:
thanks! looks good to me.

PR #403, opened 2019-04-09
get rid of soon-deprecated Cstruct.set_len, use Cstruct.sub instead

hannesm:
see https://github.com/mirage/ocaml-cstruct/pull/251 for further information


hannesm:
anyone could do me a favour and review this @mirage/core

mor1:
LGTM, though it's a while since I've looked at this in detail.

PR #409, opened 2019-07-06
ensure Xen bindings are built with right mirage-xen-ocaml CFLAGS

avsm:
Without this, the Xen checksum bindings are built with the host
CFLAGS and ABI destruction commences. Usually results in a page
fault in the Xen backend at boot time.  Fix incoming to opam-repository
as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+### dev
+* opam: ensure Xen bindings are built with right mirage-xen-ocaml CFLAGS (@avsm)
+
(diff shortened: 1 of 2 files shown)


hannesm:
I don't fully understand this approach -- this is different from e.g. https://github.com/inhabitedtype/bigstringaf - is this intentional?

avsm:
This is the same approach used by io-page-xen and mirage-xen and was just fixing an urgent regression. All of these will migrate to something that doesn't require external pkg-config variants as a separate PR :-)

PR #411, opened 2019-07-13
support ipaddr 4.0 interfaces, use Macaddr_cstruct

avsm:


src/ipv6/ndpv6.ml
```diff
@@ -110,16 +110,10 @@ let ipaddr_to_cstruct_raw i cs off =
...
 
 let macaddr_to_cstruct_raw x cs off =
-  Cstruct.blit_from_string (Macaddr.to_bytes x) 0 cs off 6
-
-let macaddr_of_cstruct cs =
-  if Cstruct.len cs <> 6 then invalid_arg "macaddr_of_cstruct";
-  match Macaddr.of_bytes (Cstruct.to_string cs) with
-  | Ok x -> x
-  | Error _ -> assert false
+  Cstruct.blit_from_string (Macaddr.to_octets x) 0 cs off 6
```
hannesm: isn't this what macaddr-cstruct should provide?

avsm: Yeah, but it doesn't right now...

yomimono:
With the right pins, this passes tests, so I'll merge and cut a release to unblock use of ipaddr/macaddr v4.

PR #415, opened 2019-07-31
introduce ipv4/Fragments.fragment : ~mtu:int -> Ipv4_packet.t -> Cstr…

hannesm:
…uct.t -> Cstruct.t list

a function that takes the MTU, an IPv4 header, and a payload, and returns a list
of IPv4 segments to be sent out. This is to-be-called with the remaining payload
after the first segment. Use this function in Static_ipv4.write.

For what its worth: >95% of IPv4 packets are not fragmented, i.e. it is crucial
that the fast path (no fragmentation) does least allocation!
In contrast to earlier code, this now copies the payload (Cstruct.concat rest)
before passing to Fragments.fragment (which copies once more
(Cstruct.append hdr payload), and then the call to writeout blits it into the
network-device allocated storage).

It is also slightly incorrect (in the more general setting where
Fragments.fragment is used from elsewhere with an Ipv4_packet.t that contains
options (which shouldn't be copied)) (requires a different representation of
options -- a list of (bool * Cstruct.t) where the bool is the first bit
(indicating whether to copy the option (or not) into following IPv4 fragments)).

//cc @yomimono @linse @cfcs
https://github.com/mirage/mirage-nat/issues/31
https://github.com/mirage/qubes-mirage-firewall/issues/73

### DIFF
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -214,7 +214,11 @@ let fragment ~mtu hdr payload =
     let opt_size = (Cstruct.len hdr.Ipv4_packet.options + 3) / 4 * 4 in
     opt_size + Ipv4_wire.sizeof_ipv4
   in
-  let data_size = mtu - hdr_size in
-  assert (data_size mod 8 = 0);
-  assert (data_size > 0);
-  List.rev (frag1 [] hdr (Cstruct.create hdr_size) data_size data_size payload)
+  let data_size =
+    let full = mtu - hdr_size in
+    (full / 8) * 8
+  in
+  if data_size <= 0 then
+    []
+  else
+    List.rev (frag1 [] hdr (Cstruct.create hdr_size) data_size data_size payload)
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -180,3 +180,38 @@ let process cache ts (packet : Ipv4_packet.t) payload =
           | Error Hole -> maybe_add_to_cache cache', None
         else
           maybe_add_to_cache cache', None
+
+(* TODO hdr.options is a Cstruct.t atm, but instead we need to parse all the
+   options, and distinguish based on the first bit -- only these with the bit
+   set should be copied into all fragments (see RFC 791) *)
+let fragment ~mtu hdr payload =
+  let rec frag1 acc hdr hdr_buf offset data_size payload =
+    let more = Cstruct.len payload > data_size in
+    let hdr' =
+      let off = offset / 8 lor (if more then 0x2000 else 0) in
+      { hdr with Ipv4_packet.off }
+    in
+    let this_payload, rest =
+      if more then Cstruct.split payload data_size else payload, Cstruct.empty
+    in
+    let payload_len = Cstruct.len this_payload in
+    Ipv4_wire.set_ipv4_csum hdr_buf 0;
+    (match Ipv4_packet.Marshal.into_cstruct ~payload_len hdr' hdr_buf with
+     | Error msg -> invalid_arg msg
+     | Ok () -> ());
+    let acc' = Cstruct.append hdr_buf this_payload :: acc in
+    if more then
+      let offset = offset + data_size in
+      (frag1[@tailcall]) acc' hdr hdr_buf offset data_size rest
+    else
+      acc'
+  in
+  let hdr_size =
+    (* padded to 4 byte boundary *)
+    let opt_size = (Cstruct.len hdr.Ipv4_packet.options + 3) / 4 * 4 in
+    opt_size + Ipv4_wire.sizeof_ipv4
+  in
+  let data_size = mtu - hdr_size in
+  assert (data_size mod 8 = 0);
+  assert (data_size > 0);
+  List.rev (frag1 [] hdr (Cstruct.create hdr_size) data_size data_size payload)
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -183,12 +183,13 @@ let process cache ts (packet : Ipv4_packet.t) payload =
 
 (* TODO hdr.options is a Cstruct.t atm, but instead we need to parse all the
    options, and distinguish based on the first bit -- only these with the bit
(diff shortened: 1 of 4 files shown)


src/ipv4/fragments.ml
```diff
@@ -180,3 +180,38 @@ let process cache ts (packet : Ipv4_packet.t) payload =
           | Error Hole -> maybe_add_to_cache cache', None
         else
           maybe_add_to_cache cache', None
+
+(* TODO hdr.options is a Cstruct.t atm, but instead we need to parse all the
+   options, and distinguish based on the first bit -- only these with the bit
+   set should be copied into all fragments (see RFC 791) *)
```
cfcs: Would it make sense with a more specific reference? RFC 791 is kind of long, and shows up with a bunch of "Updated by" and "Errata Exist".

src/ipv4/fragments.ml
```diff
@@ -180,3 +180,38 @@ let process cache ts (packet : Ipv4_packet.t) payload =
...
+      let off = offset / 8 lor (if more then 0x2000 else 0) in
+      { hdr with Ipv4_packet.off }
+    in
+    let this_payload, rest =
+      if more then Cstruct.split payload data_size else payload, Cstruct.empty
+    in
+    let payload_len = Cstruct.len this_payload in
+    Ipv4_wire.set_ipv4_csum hdr_buf 0;
+    (match Ipv4_packet.Marshal.into_cstruct ~payload_len hdr' hdr_buf with
+     | Error msg -> invalid_arg msg
```
cfcs: According to the mli docstring, this `Error msg` is returned when `hdr_buf` is "too small to represent `hdr'`. Not clear from that comment whether `~payload_len` factors in, or if that's just used to encode the expected payload length, but I assume the latter.
Do we expect this error to occur?
- If yes: Throwing an exception doesn't seem so nice :-(
- If no: Would be good to have an explanation why not (e.g. *"cannot happen because hdr_buf is always hdr_size long"*)

src/ipv4/static_ipv4.ml
```diff
@@ -64,90 +64,72 @@ module Make (R: Mirage_random.C) (C: Mirage_clock.MCLOCK) (Ethernet: Mirage_prot
...
           (* need to copy the given payload *)
-          let len, leftover =
+          let len, rest =
             Cstruct.fillv ~src:bufs ~dst:(Cstruct.shift payload_buf header_len)
           in
-          if leftover <> [] then begin
-            Log.err (fun m -> m "there's some leftover data") ;
-            invalid_arg "leftover data"
-          end ;
+          leftover := Cstruct.concat rest;
```
cfcs: Hmm this seems a bit expensive, would it be problematic to have `leftover` be a list of `Cstruct.t` to avoid having to allocate this for each fragment?

hannesm: I suspect this is covered by the commit message. yes, a performance improvement would be to redefine `Fragments.fragment` to copy less. no, I don't see a specific reason to do this (unless I see numbers showing that this is actually a bottleneck -- as mentioned earlier, most packets are not fragmented.)

src/ipv4/fragments.ml
hannesm: this has been in the code before, and indeed it is the case that `hdr_buf` was allocated in this function with `hdr_size` as argument, so it will never fail.

hannesm: I added a comment

src/ipv4/fragments.ml
```diff
@@ -180,3 +180,38 @@ let process cache ts (packet : Ipv4_packet.t) payload =
...
           maybe_add_to_cache cache', None
+
+(* TODO hdr.options is a Cstruct.t atm, but instead we need to parse all the
+   options, and distinguish based on the first bit -- only these with the bit
+   set should be copied into all fragments (see RFC 791) *)
+let fragment ~mtu hdr payload =
+  let rec frag1 acc hdr hdr_buf offset data_size payload =
+    let more = Cstruct.len payload > data_size in
+    let hdr' =
+      let off = offset / 8 lor (if more then 0x2000 else 0) in
```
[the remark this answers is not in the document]
hannesm: i added a comment and put parens around `offset / 8`.

src/ipv4/fragments.ml
hannesm: tbh, if modifying this code it is best to read 791 _and_ _all_ updates _and_ _all_ errata. I put the section 3.1 in the comment, and now refer to page 15 which includes the description of `copied`.

src/ipv4/fragments.ml
```diff
@@ -180,3 +180,38 @@ let process cache ts (packet : Ipv4_packet.t) payload =
...
+  in
+  let hdr_size =
+    (* padded to 4 byte boundary *)
+    let opt_size = (Cstruct.len hdr.Ipv4_packet.options + 3) / 4 * 4 in
+    opt_size + Ipv4_wire.sizeof_ipv4
+  in
+  let data_size = mtu - hdr_size in
+  assert (data_size mod 8 = 0);
+  assert (data_size > 0);
+  List.rev (frag1 [] hdr (Cstruct.create hdr_size) data_size data_size payload)
```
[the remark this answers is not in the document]
hannesm: wontfix

hannesm:
it would be nice to merge and releaase. One question I would like to hear opinions from others:
```OCaml
  assert (data_size mod 8 = 0);
  assert (data_size > 0);
```
is this a good way to formulate the invariants of the input parameters? should it instead be not checked (the latter - from RFC1122 3.3.2 "EMTU_R MUST be greater than or equal to 576", but I couldn't find any location checking this requirement) or computed (the former - using (`data_size / 8 * 8`) instead)?

an alternative may be to use `result` type (and return an error), but this feels overengineered (and is rather intricate to use on the callee site). WDYT?

hannesm:
removed the assert in a subsequent commit, and added docstrings to the Fragments module

avsm:
This looks fine to me. If no objections from @mirage/core I'll release on Monday.

hannesm:
thanks for your approval @avsm, merging and releasing.

PR #416, opened 2019-09-27
Add ttl parameter to Icmp.write and Udp.write

phaer:
Hi,

This MR would adapt the ICMP implementation to the proposed interface change in mirage-protocols,  see https://github.com/mirage/mirage-protocols/pull/21


hannesm:
thanks, this has been included and merged as #419

PR #417, opened 2019-10-11
opam: remove the 'build' directive on dune dependency

CraigFe:
This directive results in failure when downgrading Dune versions, due to version-specific functionality in the Dune language. See https://github.com/ocaml/opam/issues/3850 for more details.


hannesm:
thanks, CI failures are independent of this PR

PR #418, opened 2019-10-13
Ipv4.Fragments: use a mutable LRU cache (Lru.M.t instead of mutable cache : Lru.F.t)

hannesm:
- is faster (O(1) instead of O(log n))
- still avoids (due to randomization) collision-based complexity

as suggested by @pqwy in https://github.com/mirage/mirage-nat/pull/28#issuecomment-482619504 (see as well the brief overview of Lru at https://pqwy.github.io/lru/doc/lru/Lru/index.html)

it also simplifies the API of the Fragments module, and makes it easier to use! :)

### DIFF
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -58,16 +58,12 @@ end
 module K = struct
   type t = Ipaddr.V4.t * Ipaddr.V4.t * int * int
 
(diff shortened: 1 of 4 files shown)


hannesm:
CI failures are due to old (gcc < 4.9) linux distributions trying (and failing) to compile solo5.

PR #423, opened 2020-02-07
Revert "Ipv4.Fragments: use a mutable LRU cache (Lru.M.t instead of mutable cache : Lru.F.t):"

hannesm:
This reverts commit 220122426350ed5ec8fbfc139b63a1cf14dc6727.

The motivation behind this change is that Lru.M.t uses a hashtable, pre-allocated in the size of the capacity (256 * 1024 ~> 2MB just for the keys (stored in an array)). This leads to quite some memory usage (as observed in https://github.com/mirage/qubes-mirage-firewall/issues/93).

But the issue is slightly more complex: an OCaml hashtable has _amortized_ time complexity of O(1), but in worst cases it may be much bigger (adding an element may lead to resize, which means allocation of a fresh array, re-hashing of all keys, and copying) -- leading to unpredictable behaviour, which is not what we want in real systems. Also, a hashtable is never shrinking its array.

I'm fine to revert this change once we find the performance bottleneck being in the IPv4 fragment cache.

//cc @talex5

### DIFF
--- a/src/ipv4/fragments.ml
+++ b/src/ipv4/fragments.ml
@@ -58,12 +58,16 @@ end
 module K = struct
   type t = Ipaddr.V4.t * Ipaddr.V4.t * int * int
 
-  let equal a b = a = b
-
-  let hash r k = Hashtbl.seeded_hash r k
+  let compare (src, dst, proto, id) (src', dst', proto', id') =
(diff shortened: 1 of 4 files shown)


xaki23:
build works, can not test functionality right now because it makes mirage-nat fail to build.

Issue #392, opened 2019-02-03
Should q.xmit return value be ignored?


emillon:
_Originally posted by @cfcs in https://github.com/mirage/mirage-tcpip/pull/370#issuecomment-443269244_

@avsm Do you remember what this function (`xmit`) actually does?

@hannesm suggested that we could make it log when we get a `Error _`, but I can't figure out if this is supposed to fail, and I felt reluctant to add it in case it was supposed to fail 10k times per second in some cases. If it is not supposed to fail often, I think adding a warning message to the log sounds like a very sensible suggestion, especially if it is supposed to *never* fail and there are bugs to be fixed hiding in here.

yomimono:
`xmit` will be some invocation of `WIRE.xmit` which is [this function](https://github.com/mirage/mirage-tcpip/blob/v3.7.1/src/tcp/wire.ml#L48).  It should fail only if the buffer given is too small or if the underlying interface fails to write the IP packet, if I'm reading correctly.

PR #425, opened 2020-02-07
tcp: ignore result of xmit, continues #310, see #392

hannesm:
CI is failing since lwt >= 5 is installed, but the TCP code is not yet ready for this. #310 had a partial fix (`Lwt.async` now demands a `unit Lwt.t` result type), while this applies the same patch to the other call site of `xmit`.


hannesm:
this PR establishes compatibility with lwt >= 5 :D

PR #426, opened 2020-06-17
adapt to ipaddr 5.0.0 changes: Static_ipv4.connect only needs a Ipaddr.V4.Prefix.t

hannesm:


hannesm:
I'd like to merge and release this (as 5.0.0) unless someone objects. The CI is failing since the signature of Static_ipv4.connect has changed, and the mirage tool needs to be updated with this change. That's the reason why it should be 5.0.0 -- so that older mirage tools that emit the `connect` calls do not use the new code (with a modified interface).

PR #429, opened 2020-08-25
stack-direct & tcp: implement disconnect

hannesm:
- TCP close()s all active channels and avoids establishing new connections
- Stack-direct cancels the listener task

An example use side will look like:
```OCaml
Tcp.disconnect tcp >>= fun () ->
Stack_direct.disconnect stack >>= fun () ->
(* now ready for re-setup of a stack, i.e. IP configuration changed *)
```


MagnusS (approved):

hannesm:
I removed the sleep. The behaviour of Linux and FreeBSD is:
- send out a FIN+ACK for existing TCP connections
- when the other side replies, send a RST

That's what this PR does as well now.

hannesm:
all green and approved -- merging

PR #430, opened 2020-08-26
IPv4 routing: treat "Ipaddr.V4.Prefix.broadcast network" also as broadcast address

### LINKED ISSUE #427: creating broadcast in unikernels
There is a remote server that will send broadcasts and there are multiple unikernels listening to the broadcast. How do we create this broadcast address

hannesm:
fixes #427

### DIFF
--- a/src/ipv4/routing.ml
+++ b/src/ipv4/routing.ml
@@ -17,7 +17,9 @@ module Make(Log : Logs.LOG) (A : Mirage_protocols.ARP) = struct
   open Lwt.Infix
 
   let destination_mac network gateway arp = function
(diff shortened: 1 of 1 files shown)


MagnusS (approved):

hannesm:
on the receiving path, this is already handled similarly:
```OCaml
      let of_interest ip =
        Ipaddr.V4.(compare ip (Prefix.address t.cidr) = 0
                   || compare ip broadcast = 0
                   || compare ip (Prefix.broadcast t.cidr) = 0)
      in
      if not (of_interest packet.dst) then error else deal_with_it
```

PR #431, opened 2020-08-27
IPv6 stack support

MagnusS:
This adds support for configuring a V6 stack in tcpip_stack_direct and adds a TCPv6 connect test that succeeds. ~~It depends on unmerged changes in mirage-stack so shouldn't be merged yet (mirage-stack is pinned to my branch).~~

~~I had to disable the test that adds trailing bytes to the ethernet frames, as this results in a lot of packets being dropped due to checksum errors in our stack (checksum in pcap is fine). This seems to indicate that the implementation calculates the checksum of the full buffer - not just the IP packet length, but I haven't investigated this further.~~ (fixed)

The stack will also queue packets without a source IP before the IPv6 address has been negotiated. These packets will be sent when the IP is available and confuse the other end. I've added a temporary fix to wait for the IPv6 address to be configured before the test starts.

### DIFF
--- a/test/vnetif_common.ml
+++ b/test/vnetif_common.ml
@@ -34,6 +34,7 @@ sig
   type 'a io
   type id
   module Stackv4 : Mirage_stack.V4
+  module Stackv6 : Mirage_stack.V6
 
   (** Create a new backend *)
   val create_backend : unit -> backend
@@ -44,6 +45,10 @@ sig
 
   (** [create_stack backend ?mtu ip netmask gateway] adds a listener
       function to the backend *)
+  val create_stack_v6 : ?mtu:int -> ?ip:Ipaddr.V6.t list ->
+    ?netmask:Ipaddr.V6.Prefix.t list ->
+    ?gateways:Ipaddr.V6.t list -> backend -> Stackv6.t Lwt.t
+
   val create_backend_listener : backend -> (buffer -> unit io) -> id
 
   (** Disable a listener function *)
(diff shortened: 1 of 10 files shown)


test/vnetif_common.ml
```diff
@@ -64,28 +69,56 @@ struct
...
+  let rec wait_for_ipv6 t () =
+    (* Wait for IP to be valid *)
+    if List.length (Ip6.get_ip (Stackv6.ip t)) >= 1 then
+    begin
+            Lwt.return t
+    end else
+    Time.sleep_ns (Duration.of_ms 50) >>= fun () ->
+    wait_for_ipv6 t ()
+
   let create_stack ?mtu ~cidr ?gateway backend =
```
dinosaure, line 101: May be it's worth to make an alias `create_stack_v4`?

dinosaure: nvm, it's about tests so we don't really care :+1:

dinosaure (approved):

hannesm:
thanks @MagnusS 

> I had to disable the test that adds trailing bytes to the ethernet frames, as this results in a lot of packets being dropped due to checksum errors in our stack (checksum in pcap is fine). This seems to indicate that the implementation calculates the checksum of the full buffer - not just the IP packet length, but I haven't investigated this further.

I fixed the code in 6d68284, which includes some more safety (i.e. avoiding Cstruct exceptions in get_yyy).

MagnusS:
Thanks for the fixes @hannesm !

Now that ipv6 generally seems to work I also enabled the iperf test to test TCP connections over IPv6. I've tested with ~100mb transfers over vnetif and it seems to work well, also with packet loss (at least as long as both ends are mirage unikernels...).

The iperf test now also reports the results every second with `Logs.info`, so it's possible to run an ipv4/ipv6 comparison by executing test.exe directly: `ALCOTEST_VERBOSE=1 dune exec test/test.exe test iperf 6` (a bug had been introduced at some point which made the output kbit/ns and very noisy)

On my machine it looks like the ipv4 test is a little bit faster with an avg rate of ~657 Mbit/s, while IPv6 is ~627 Mbit/s. This is measured over 10 x `test/test.exe test iperf 6`, which will transfer around 100mb of data. This seems very reasonable as we just started using this part of the stack.

hannesm:
I'm fine with these changes, but would wait until https://github.com/ocaml/opam-repository/pull/17090 hits opam-repository, and then remove the mirage-stack pin (and instead put a lower bound on `"mirage-stack" {>= "2.1.0"}` into tcpip.opam) before merging this PR.

MagnusS:
mirage-stack 2.1.0 is merged -- I rebased and updated `tcpip.opam`. I think this is ready to be merged now when CI is green

PR #433, opened 2020-09-11
move Tcpip_stack_socket to Tcpip_stack_socket.V4, provide Tcpip_stack_socket.V6; also dual direct stack and fixes

hannesm:
..continuing the #431 ride.. //cc @MagnusS

### DIFF
--- a/src/stack-direct/tcpip_stack_direct.ml
+++ b/src/stack-direct/tcpip_stack_direct.ml
@@ -376,9 +376,9 @@ module IPV4V6 (Ipv4 : Mirage_protocols.IPV4) (Ipv6 : Mirage_protocols.IPV6) = st
     List.map (fun ip -> Ipaddr.V4 ip) (Ipv4.get_ip t.ipv4) @
     List.map (fun ip -> Ipaddr.V6 ip) (Ipv6.get_ip t.ipv6)
 
-  let mtu t =
-    (* TODO incorrect for IPv4 *)
-    Ipv6.mtu t.ipv6
+  let mtu t ~dst = match dst with
+    | Ipaddr.V4 dst -> Ipv4.mtu t.ipv4 ~dst
+    | Ipaddr.V6 dst -> Ipv6.mtu t.ipv6 ~dst
 end
 
 module MakeV4V6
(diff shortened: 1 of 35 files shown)


hannesm:
travis is failing since this is an API-breaking change which needs updates to the mirage tool (there are some needed for #432 as well). i'd suggest to merge #432 and #433 (and an upcoming #434), cut a release (6.0.0 -- to avoid earlier mirage releases using the new, API-incompatible tcpip), and cut a mirage release (3.8.1).

hannesm:
this now requires https://github.com/mirage/mirage-protocols/pull/27 and https://github.com/mirage/mirage-stack/pull/19, and fix some bugs:
- UDP write now has a `?src` argument (crucial for IPv6 where we deal with at least two IP addresses)
- TCP write uses its `src` for checksum computation and passing it onto the IP layer (otherwise, the other peer may receive a reply from an unexpected IP address)
- Direct dual IPv4 and IPv6 stack implementation

hannesm:
moving forward, I first cut a 5.0.1 release (based on master, but reverting the API-breaking change of Ipv6.connect -- which mirage knows about). now we can figure the IPv6 out subsequently, break APIs and move to a 6.0.0 release. I pushed the 5.0.1 changes entry to master directly.

hannesm:
now with `setsockopt IPV6_ONLY true/false` there's (a) the IPv6 socket stack doing _only_ ipv6, and a dual socket stack. I'll revise together with the mirage PR (https://github.com/mirage/mirage/pull/1187) for some more convenient usage.

src/stack-direct/tcpip_stack_direct.ml
```diff
@@ -256,3 +256,236 @@ module MakeV6
...
+    | Ipaddr.V4 dst -> Ipaddr.V4 (Ipv4.src t.ipv4 ~dst)
+    | Ipaddr.V6 dst -> Ipaddr.V6 (Ipv6.src t.ipv6 ~dst)
+
+  let get_ip t =
+    List.map (fun ip -> Ipaddr.V4 ip) (Ipv4.get_ip t.ipv4) @
+    List.map (fun ip -> Ipaddr.V6 ip) (Ipv6.get_ip t.ipv6)
+
+  let mtu t =
+    (* TODO incorrect for IPv4 *)
+    Ipv6.mtu t.ipv6
```
hannesm, line 381: this should be addressed before a merge and release -- current idea is to adapt on the IP layer the mtu function from `t -> int` to `t -> dst:ipaddr -> int` - which allows to inspect the IP protocol version. requires a change in mirage-protocols though (but we're bumping its major anyways for other required changes)/

hannesm: this has been done and incorporated into the mirage-protocols 5.0.0 release.

PR #436, opened 2020-11-25
Migrate to dune 2.7 and add bisect runes

### LINKED ISSUE #160: use bisect to test coverage of the tcpip tests
instructions from @djs55 on the mailing list below:

---

I've been having a lot of fun with bisect[1] recently-- I've been using it to see which parts of the mirage-block-volume[2] and shared-block-ring[3] libraries are completely untouched by their unit tests. I found the following game to be quite addictive:
1. "make coverage", and load result in web-browser
2. spot a big chunk of obvious red (danger, danger, danger)
3. (thinking carefully about what could go wrong) devise an interesting test to stress the red bits (obviously you could cover it with a 'noddy' test but there is probably no point)
4. "make coverage", reload in browser and see the red go green!

I've hooked it up with coveralls.io (admittedly in a bit of a hacky way[4]) such that the master branch is firmly in "development" mode, linking against bisect, and without checking in the oasis autogen rubbish. The .travis.yml runs both the travisci-skeleton script and then invokes ocveralls[5] to upload the results.

I've added a separate "make release" step which removes bisect and checks in the autogen (presumably into a release branch). Perhaps eventually this could make a github pull request (with the "hub" tool?) and make an opam package?

I think the game is made even more addictive when the coveralls badge changes colour, see:

https://coveralls.io/r/mirage/mirage-block-volume?branch=master

Here's an example bisect report (the code is a work-in-progress):

http://dave.recoil.org/tmp/report/file0000.html

If you haven't given bisect a go -- I recommend playing with it.

Also, if you can think of a nicer way to integrate this -- I admit using 'sed' on the _oasis file is a bit of a hack -- please let me know!

Cheers,
Dave

[1] http://bisect.x9c.fr
[2] https://github.com/mirage/mirage-block-volume
[3] https://github.com/mirage/shared-block-ring
[4] https://github.com/mirage/shared-block-ring/commit/67b9f3100be8e4e9732dd79b7c1cc5352a61d478
[5] https://github.com/sagotch/ocveralls

hannesm:
this fixes #160 //cc @avsm

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,2 @@
+version = 0.15.0
+disable = true
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,2 @@
-(lang dune 1.0)
+(lang dune 2.7)
 (name tcpip)
--- a/examples/ping/dune
+++ b/examples/ping/dune
@@ -1,3 +1,3 @@
 (executables
  (names ping)
- (libraries cmdliner logs logs.fmt tcpip.icmpv4-socket))
+ (libraries cmdliner logs logs.fmt tcpip.icmpv4-socket tcpip.unix))
--- a/src/icmp/dune
+++ b/src/icmp/dune
@@ -2,5 +2,6 @@
  (name tcpip_icmpv4)
  (public_name tcpip.icmpv4)
  (libraries mirage-protocols rresult logs tcpip mirage-profile tcpip.udp)
- (preprocess (pps ppx_cstruct))
+ (preprocess
+  (pps ppx_cstruct))
  (wrapped false))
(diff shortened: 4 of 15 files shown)


hannesm:
@MagnusS do you have an idea why the "iperf with two stacks and uniform packet loss of packets with no payload" fails with:
```
ASSERT iperf test timed out after 120.000000 seconds
test.exe: [INFO] listen of mac=02:50:00:00:00:02,ip=10.0.0.100 cancelled
FAIL iperf test timed out after 120.000000 seconds
```

i.e. running in a timeout, esp. on the arm32 runner -- should we increase the timeout? (it is 120 seconds atm)

hannesm:
(increasing the timeout for iperf did the trick, I disabled the ipv6 iperf test -- marked it as slow)

hannesm:
anyone would like to review this? otherwise I'll go ahead merge, release, and fix up fallout

PR #441, opened 2021-02-18
Use the usual layout to compile freestanding C stubs

dinosaure:
This PR wants to provide a smooth transition between MirageOS 3 and MirageOS 4. Even if the goal permits a long term support of MirageOS 3 and `tcpip`, the real goal is to avoid any constraints between `tcpip`, MirageOS 3 and MirageOS 4.

As we checked (@TheLortex and me) the usual layout for C stubs available on `mirage-crypto` and `digestif` is compatible with MirageOS 4. `tcpip` (due to the story) still continue to use some corner cases about the MirageOS 3 linking step. This PR is like an unification of C stubs in our MirageOS eco-system.

By this way and if we cut a release with that, we don't need to upgrade `tcpip` for MirageOS 4 and we don't need to care about incompatible details between MirageOS 3, MirageOS 4 and this version of `tcpip`. So it paves the way for a smooth transition.

**NOTE**: the new artifact `libtcpip_freestanding_stubs.a` is needed only for MirageOS 3 (at the linking step). Then, when time forgets MirageOS 3, we will be able to delete that.


hannesm:
AFAICT at the moment the C stubs are provided by mirage-xen & mirage-solo5 -- including them here will lead to duplicated symbols at the link step (this means, mirage-solo5/mirage-xen should be updated to not provide the symbols anymore at the same time)... maybe this PR should be delayed until mirage4 is out (which, IIUC, will handle C stubs without the need for extra Makefiles)?

dinosaure:
> AFAICT at the moment the C stubs are provided by mirage-xen & mirage-solo5 -- including them here will lead to duplicated symbols at the link step (this means, mirage-solo5/mirage-xen should be updated to not provide the symbols anymore at the same time)...

It seems that the `gcc` linker is less strict because, with this PR, I can link the unikernel.

Currently, `tcpip` is incompatible with MirageOS 4 due to this old linking-trick - next week, I will try to figure out why precisely.  This PR is only a bridge between MirageOS 3 and MirageOS 4 and deserve only one purpose: with that, we don't need to cut a release of `tcpip` after MirageOS 4 and we don't need to apply constraints.

However, as you can say, it seems that, at least for FreeBSD I believe, this patch does not work.

So, to clear the situation, without any solution:
1) when MirageOS 4 will be released it will be incompatible with this version of `tcpip`
2) we need to make a release of `tcpip` to be able compatible with MirageOS 4 **and** incompatible with MirageOS 3

For my perspective, may be it's the time to delete `checksum` stubs on `mirage-solo5`/`mirage-xen`?

hannesm:
to avoid conflicts / unnecessary churn, I propose:
- (a) rename the C symbols used by tcpip_checksum and add the freestanding "cross-compilation" runes, release tcpip (this will now work with mirage 3 and mirage 4)
- (b) remove the old checksum stubs from mirage-solo5 and mirage-xen

The renaming leads to old mirage-solo5 and new tcpip continues to be symbol-conflict-free linkable (old tcpip releases will need to conflict with new mirage-solo5/mirage-xen (those with the checksum_stubs removed)).

overall, it is not clear to me what the impact is: what about cstruct stubs and io-page stubs? anything else that is now in mirage-solo5/mirage-xen, and should instead be in the respective library that uses it?

dinosaure:
> overall, it is not clear to me what the impact is: what about cstruct stubs and io-page stubs? anything else that is now in mirage-solo5/mirage-xen, and should instead be in the respective library that uses it?

We did not find a problem for `cstruct` mostly because `mirage-tcpip` has a sub-package `tcpip.unix` which takes C stubs - otherwise, you have missing symbols for a simple executable. This is not the case for `cstruct` which takes C stubs by default and I believe that `io-pages` has some trouble.

dinosaure:
I prepended C functions with `mirage_`, I can take care about the release. As far as I can tell, a minor one should be enough.

hannesm:
to me this change looks good. should we move forward and remove the tcpip.unix package entirely? (this would be a breaking change, but I think it is fine - there are barely users of tcpip.unix AFAICT)

hannesm:
I looked, and `tcpip.unix` is hardcoded e.g. in the mirage tool. While with this PR this sublibrary does not serve any purpose anymore, let's keep it until the next major release (or until we have a mirage released that does not depend on `tcpip.unix`).

hannesm (approved):
LGTM. Please merge and release when CI is happy.

Issue #437, opened 2021-01-19
Be able to properly close the underlying socket for a server


dinosaure:
With `tcpip.unix`, when we start a TCP/IP server on a specific port and we want to delete it (like `Lwt_unix.close`), the current implementation does not properly do that. Then, if we start a new TCP/IP service on the same address/port, we got an `EADDRINUSE` because the underlying old (closed?) socket still is bound on the same `addr:port`.

For the use case, if we want to download a let's encrypt certificate (and launch a simple HTTP server to do the challenge) and start then our proper HTTP + HTTPS server, we got this error with `mirage configure -t unix`.

dinosaure:
Close by #443

PR #443, opened 2021-03-22
Add cancelation on tcpip.stack-socket

dinosaure:
This PR wants to fix an issue about cancelation of listening socket. You can see the error on this CI output:
https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/661b80c8d56556cd711fb2e8f7a1c0edae56bf15

I would like to have a review to incorporate it into `6.1.0` and it should close the issue #437. (/cc @talex5 & @hannesm)

### DIFF
--- a/src/stack-unix/tcpip_stack_socket.ml
+++ b/src/stack-unix/tcpip_stack_socket.ml
@@ -97,7 +97,7 @@ module V4 = struct
           in
           Lwt.catch loop (fun _-> Lwt.return_unit) >>= fun () -> Lwt_unix.close fd)
 
-  let listen t = t.switched_off >>= fun () -> Lwt.return_unit
+  let listen t = t.switched_off
 
   let connect udpv4 tcpv4 =
     Log.info (fun f -> f "IPv4 socket stack: connect");
--- a/src/stack-unix/tcpip_stack_socket.ml
+++ b/src/stack-unix/tcpip_stack_socket.ml
@@ -27,8 +27,8 @@ module V4 = struct
   type t = {
     udpv4 : UDPV4.t;
     tcpv4 : TCPV4.t;
-    stop : [ `Stopped ] Lwt.u;
-    switched_off : [ `Stopped ] Lwt.t;
+    stop : unit Lwt.u;
+    switched_off : unit Lwt.t;
   }
 
   let udpv4 { udpv4; _ } = udpv4
@@ -46,7 +46,7 @@ module V4 = struct
           UDPV4.get_udpv4_listening_fd t.udpv4 port >>= fun fd ->
           let buf = Cstruct.create 4096 in
           let rec loop () =
-            (* TODO cancellation *)
+            if not (Lwt.is_sleeping t.switched_off) then raise Lwt.Canceled ;
             Lwt.catch (fun () ->
                 Lwt_cstruct.recvfrom fd buf [] >>= fun (len, sa) ->
                 let buf = Cstruct.sub buf 0 len in
@@ -61,7 +61,8 @@ module V4 = struct
                  Lwt.return_unit) >>= fun () ->
             loop ()
           in
-          Lwt.pick [ t.switched_off; loop () ] >>= fun `Stopped -> Lwt_unix.close fd)
+          Lwt.catch loop (fun _ -> Lwt.return_unit) >>= fun () ->
+          Lwt_unix.close fd)
 
   let listen_tcpv4 ?keepalive t ~port callback =
     if port < 0 || port > 65535 then
@@ -76,6 +77,7 @@ module V4 = struct
           (* TODO cancellation *)
           let rec loop () =
             Lwt.catch (fun () ->
+                if not (Lwt.is_sleeping t.switched_off) then raise Lwt.Canceled ;
                 Lwt_unix.accept fd >|= fun (afd, _) ->
                 (match keepalive with
                  | None -> ()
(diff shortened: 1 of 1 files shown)


hannesm:
I'm not sure I understand the patch. What is the goal in semantics that you try to achieve?

Why are there some changes to `V4.listen_tcpv4` (but none to `listen_udpv4`)? What about the V6 and the V4V6 modules in the same file?

Is it possible to formulate a test case with the desired semantics (to see what is failing at the moment, and what the intention is)?

hannesm:
What is the advantage of using a `Lwt_switch.t` here? Wouldn't a `Lwt_condition.t` be sufficient?

dinosaure:
> Why are there some changes to V4.listen_tcpv4 (but none to listen_udpv4)? What about the V6 and the V4V6 modules in the same file?

Yes, I forget to apply such change on all stacks. Now it's fixed by a forced-push.

> I'm not sure I understand the patch. What is the goal in semantics that you try to achieve?

Currently, the _master_ socket used to listen is not properly closed which leads an error when you want to launch multiple times sequentially a server on a specific port. Such situation mostly appears when we do some tests and I got several times the `EADDRINUSE`. The reproducibility of such situation is hard and seems intrinsic to the underlying operating system used and it seems the case on the `opam-ci`.

> What is the advantage of using a Lwt_switch.t here? Wouldn't a Lwt_condition.t be sufficient?

This is where @talex5 should have a better explanation than me.

src/stack-unix/tcpip_stack_socket.ml
```diff
@@ -59,7 +61,7 @@ module V4 = struct
                  Lwt.return_unit) >>= fun () ->
             loop ()
           in
-          loop ())
+          Lwt.pick [ t.switched_off; loop () ] >>= fun `Stopped -> Lwt_unix.close fd)
```
talex5, line 64: Note that `pick` will try to cancel `loop`, which is unreliable. If it's at `recvfrom` then it will probably work, but if it's in `callback` the cancellation might just be silently dropped. You might want to check whether `t.switched_off` is resolved each time around the loop, just before doing the `recvfrom`, to prevent that.

dinosaure: Something like:
```ocaml
    if port < 0 || port > 65535 then
      raise (Invalid_argument (err_invalid_port port))
    else
      (* FIXME: we should not ignore the result *)
      Lwt.async (fun () ->
          UDPV4.get_udpv4_listening_fd t.udpv4 port >>= fun fd ->
          let buf = Cstruct.create 4096 in
          let rec loop () =
            (* TODO cancellation *)
            Lwt.catch (fun () ->
                Lwt.pick
                  [ (t.switched_off >>= fun () -> Lwt.return `Stopped)
                  ; (Lwt_cstruct.recvfrom fd buf [] >>= fun (len, sa) ->
                     Lwt.return (`Recv (len, sa))) ] >>= function
                | `Stopped -> Lwt.return `Stopped
                | `Recv (len, sa) ->
                  let buf = Cstruct.sub buf 0 len in
                  (match sa with
                   | Lwt_unix.ADDR_INET (addr, src_port) ->
                     let src = Ipaddr_unix.V4.of_inet_addr_exn addr in
                     let dst = Ipaddr.V4.any in (* TODO *)
                     callback ~src ~dst ~src_port buf >>= fun () ->
                     Lwt.return `Continue
                   | _ -> Lwt.return `Continue))
              (fun exn ->
                 Log.warn (fun m -> m "exception %s in recvfrom" (Printexc.to_string exn)) ;
                 Lwt.return `Continue) >>= function
            | `Continue -> loop () | `Stopped -> Lwt_unix.close fd
          in
          loop ())
```

is better?

talex5: I was thinking something more like:
```ocaml
if not (Lwt.is_sleepting t.switched_off) then raise Lwt.Canceled;
Lwt.catch (fun () -> ...
```

dinosaure: Oh ok :+1:, I will apply such way on all `listen_*` functions.

src/stack-unix/tcpip_stack_socket.ml
```diff
@@ -74,6 +77,7 @@ module V4 = struct
           (* TODO cancellation *)
           let rec loop () =
             Lwt.catch (fun () ->
+                if not (Lwt.is_sleeping t.switched_off) then raise Lwt.Canceled ;
```
talex5, line 80: Why is this inside the `catch`? It will log the cancellation and accept again.

src/stack-unix/tcpip_stack_socket.ml
```diff
@@ -59,7 +61,8 @@ module V4 = struct
                  Lwt.return_unit) >>= fun () ->
             loop ()
           in
-          loop ())
+          Lwt.catch loop (fun _ -> Lwt.return_unit) >>= fun () ->
```
talex5, line 64: It would be better to ignore just the `Canceled` exception, not all of them I think.

src/stack-unix/tcpip_stack_socket.ml
```diff
@@ -91,17 +95,16 @@ module V4 = struct
                  Lwt.return_unit) >>= fun () ->
             loop ()
           in
-          loop ())
+          Lwt.catch loop (fun _-> Lwt.return_unit) >>= fun () -> Lwt_unix.close fd)
 
-  let listen _t =
-    let t, _ = Lwt.task () in
-    t (* TODO cancellation *)
+  let listen t = t.switched_off >>= fun () -> Lwt.return_unit
```
talex5, line 100:
```suggestion
  let listen t = t.switched_off
```

dinosaure:
I think we are ready for a release 👍 .

PR #447, opened 2021-03-29
Make freestanding compilation possible w/o opam

sternenseemann:
If we miss opam we just invoke pkg-config with PKG_CONFIG_PATH from the
environment (if any). Also don't override a manually passed
PKG_CONFIG_PATH even if we have opam.

Patch taken from <https://github.com/NixOS/nixpkgs/pull/117948>.

### DIFF
--- a/freestanding/Makefile
+++ b/freestanding/Makefile
@@ -1,5 +1,5 @@
 ifneq (, $(shell command -v opam))
-	PKG_CONFIG_PATH ?= $(shell opam config var prefix)/lib/pkgconfig
+	PKG_CONFIG_PATH ?= $(shell opam var prefix)/lib/pkgconfig
 endif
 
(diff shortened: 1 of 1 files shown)


freestanding/Makefile
```diff
@@ -1,4 +1,6 @@
-PKG_CONFIG_PATH := $(shell opam config var prefix)/lib/pkgconfig
+ifneq (, $(shell command -v opam))
+	PKG_CONFIG_PATH ?= $(shell opam config var prefix)/lib/pkgconfig
```
hannesm, line 2:
```suggestion
	PKG_CONFIG_PATH ?= $(shell opam var prefix)/lib/pkgconfig
```

let's proactively avoid warnings with opam 2.1 (see https://github.com/ocaml/opam/issues/4503)

sternenseemann: Let's see if CI likes it.

hannesm:
thanks, I rebased onto the current main branch.

PR #448, opened 2021-04-16
Restore compatibility with Alcotest.1.4.0

CraigFe:
The Alcotest 1.4.0 release adds optional arguments to Alcotest.fail.  Eta-expansion is necessary to use this function with the continuation printers.


hannesm:
thanks, I force-pushed to have it based on the current main branch, and also fixed a second occurrence of Alcotest.fail.

Issue #446, opened 2021-03-29
`EADDRINUSE` raised by a test on a specific platform


dinosaure:
For the release of 6.1.0, we got an error on `opam-ci` on a test about sockets:
```
# ┌──────────────────────────────────────────────────────────────────────────────┐
# │ [FAIL]        socket                0   two sockets connect via TCP.         │
# └──────────────────────────────────────────────────────────────────────────────┘
# test.exe: [INFO] IPv4 socket stack: connect
# test.exe: [INFO] IPv4 socket stack: connect
# [exception] Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
#             Raised by primitive operation at Tcpip_stack_socket.V4.listen_tcpv4 in file "src/stack-unix/tcpip_stack_socket.ml", line 77, characters 6-88
#             Called from Dune__exe__Test_socket.two_connect_tcp.(fun) in file "test/test_socket.ml", line 40, characters 2-60
#             Called from Dune__exe__Test.run in file "test/test.ml" (inlined), line 37, characters 15-24
#             Called from Dune__exe__Test.(fun) in file "test/test.ml", line 51, characters 42-47
#             Called from Alcotest_engine__Core.Make.protect_test.(fun) in file "src/alcotest-engine/core.ml", line 277, characters 17-23
#             Called from Alcotest_engine__Monad.Identity.catch in file "src/alcotest-engine/monad.ml", line 34, characters 31-35
```

Such error seems reproducible on:
- centos-8
- debian-testing
- opensuse-15.2

At this time, nobody reproduced such error locally. For more details, you can see this PR:
https://github.com/ocaml/opam-repository/pull/18357#issuecomment-801696388

MagnusS:
After #443, wouldn't the ports only close when exiting the blocking calls to recvfrom, accept? Then they may still be listening when the next test runs (even after close).

https://github.com/mirage/mirage-tcpip/blob/master/src/stack-unix/tcpip_stack_socket.ml#L83

The tests may now be failing because we missed the exception in the past:

https://github.com/mirage/mirage-tcpip/commit/d63b65db0f61db864aa8d0ff82467471656cf924#diff-cd11935cb53f2f3aa659cee9d12de7c8db4b502f94ba978a4b87d5df181dd0be

In the test we don't check that the result is read, so it could have worked before because it was sent to a different stack that was still listening:

https://github.com/mirage/mirage-tcpip/blob/master/test/test_socket.ml#L44

[...]

PR #449, opened 2021-07-14
Fix udp socket

hannesm:


hannesm:
@mirage/core this came out of using stack-socket. unfortunately SO_REUSEPORT is only available from OCaml 4.12 onwards -- anyone has problems with dropping earlier OCaml releases in tcpip?

djs55:
I'm personally happy with OCaml 4.12+

src/stack-unix/udpv4_socket.ml
```diff
@@ -30,6 +30,7 @@ let get_udpv4_listening_fd {listen_fds;interface} port =
     Lwt.return @@ Hashtbl.find listen_fds (interface,port)
   with Not_found ->
     let fd = Lwt_unix.(socket PF_INET SOCK_DGRAM 0) in
+    Lwt_unix.(setsockopt fd SO_REUSEPORT true);
```
djs55, line 33: Is this to allow multiple Unix processes to bind to the same port, to share responsibility for handling the requests?

hannesm: This is to allow the same unix process to bind to the same port several times, i.e. in DNS:
- we listen on UDP Ipaddr_any, 53 (via a listen callback) -- calls `bind(socket, *, 53)`
- once a UDP frame is received, we take some time to find an answer, and eventually send _from_ (*, 53) to the client -- this means we have to bind as well to (*, 53).

Without the REUSEPORT we can't bind to the same ip, port pair multiple times.

dinosaure:
> @mirage/core this came out of using stack-socket. unfortunately SO_REUSEPORT is only available from OCaml 4.12 onwards -- anyone has problems with dropping earlier OCaml releases in tcpip?

Hmmhmm, 4.12 is a bit higher for me. We should keep, at least, the support of the 3 last version of OCaml I think.

hannesm:
> 4.12 is a bit higher for me

ok, so what are the options?
- leave tcpip broken as is (the socket stack can't receive on the same port as it wants to send on (via UDP))
- provide a C stub that allows to set SO_REUSEPORT
- rework this PR to be able to avoid SO_REUSEPORT (the commit https://github.com/mirage/mirage-tcpip/commit/76b857821c1af5d4bfd603b0df790e04489d4159#diff-2df233b86746a505609a07125cd5a1ca625c8ddc4faba8ab44ddf9c2a0d9b578 introduced the `create_socket` asymmetry in Udpv4v6_socket) -- the `create_socket` exists since when sending UDP data we only want to use a single socket (and the get_udpv4v6_listening_fds may return two sockets (one for IPv4, one for IPv6 if both a static v4 and a v6 address is configured). Likely this can be refactored.

Eventually, the SO_REUSEPORT is not required if -- similar to how Udpv4_socket (and Udpv6_socket) re-use listening sockets for writing. Unfortunately the lifetime of a socket is then infinite (with file descriptors leakage) -- I guess the stack-socket needs some enhancements anyways (and there should be the counterpart to llsten_tcp/listen_udp that stops the listener).

> 3 last version of OCaml

From my memory, we agreed to the last 2 OCaml versions -- and 4.13 is upcoming soon. I also think that the tcpip stack is barely used by others than MirageOS unikernels. IMHO it is fine to bump to 4.12 (and 4.13) for this library.

So:
- this PR here fixes some issues if you want to use UDP and the socket stack, esp. with dual stacks
- there's more work needed to avoid leaking file descriptors
- it is not urgent since it seems like nobody really uses the socket stack and UDP (we use it for developing and testing a dns resolver, but we can work with a local pin for now)

hannesm:
I opened #450 describing the leaks and potential fix(es). related to #446. I also moved the less controversial part to #451.

dinosaure:
> From my memory, we agreed to the last 2 OCaml versions -- and 4.13 is upcoming soon. I also think that the tcpip stack is barely used by others than MirageOS unikernels. IMHO it is fine to bump to 4.12 (and 4.13) for this library.

So if we agree with that, it should fine so :+1:. I largely prefer a non-buggy stack.

djs55 (approved):

avsm:
A minimum of 4.12 is just fine by me for this package

hannesm:
we revised the implementation to address the file descriptor leaks (should solve #446 / #450).
also, we removed the REUSEPORT socket option (which was really only needed in udpv4v6_socket), and instead reuse the same socket for listen and write (by adding some more logic into get_udpv4v6_listening_fd).

TL;DR: the socket stack `disconnect` function now cleanup file descriptors. the tests suite should pass nicely, and we don't require OCaml 4.12. we tested this with the dns-resolver unikernel which reads from and writes to a lot of udp sockets.

any reviews welcome, I'm keen to merge and release a new version of tcpip without the leaks (plus the other changes in previous PRs).

PR #457, opened 2021-11-07
Adapt to mirage/mirage-protocols#28 changes:

### LINKED ISSUE #452: unlisten of a port
A `listen` (`_tcp` / `_udp`) function exists, but there's no way to not listen to that port anymore. Now, in some use cases (e.g. a DNS resolver), all that is needed is "send packet from port P to Y:P'" and "wait for an answer to port P (from Y:P')", but once the answer is received, the listening socket on port P can be safely closed.

Also note that the current semantics of calliing `listen` mutliple times with the same port on the same stack diverges between socket and direct stack -- the socket stack will fail (ADDRINUSE), the direct stack will overwrite the listener.

hannesm:
TCP and UDP layers (direct & socket):
type ipinput is gone
val handle no longer has a ~listener argument
val listen : t -> ~port:int -> (?keepalive) -> callback -> unit
val unlisten : t -> ~port:int -> unit

On the code level: the listener hash tables moved from stack-* to tcp and udp
(tcp/flow; udp/udp; tcp{v4,v6.v4v6}_socket; udp{v4,v6,v4v6}_socket).

This moved quite some code in the socket stack from the tcpip_stack_socket to
the respective tcp/udp layers.

Adapt to mirage/mirage-protocols#28 changes (that are required); fixes #452

Any review would be highly appreciated. The merge and release plan is to first cut a mirage-protocols release (with exising tcpip getting upper bounds), then a mirage-tcpip release with mirage-protocols lower bound.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### v6.4.0 (2021-11-11)
+
+* Adapt to mirage-protocols 6.0.0 API (#457 @hannesm)
+* TCP and UDP now have a listen and unlisten function (fixes #452)
+* type ipinput (in TCP and UDP) and listener (in TCP) have been removed
+
 ### v6.3.0 (2021-10-25)
 
 * Use Cstruct.length instead of deprecated Cstruct.len (#454 @hannesm)
--- a/src/stack-direct/tcpip_stack_direct.ml
+++ b/src/stack-direct/tcpip_stack_direct.ml
@@ -19,14 +19,11 @@ open Lwt.Infix
 let src = Logs.Src.create "tcpip-stack-direct" ~doc:"Pure OCaml TCP/IP stack"
 module Log = (val Logs.src_log src : Logs.LOG)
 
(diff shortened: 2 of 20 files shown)


dinosaure:
So I did a quick look on this PR (don't have enough time to really go deeper) and I really like the interface which gives to us a chance to de-allocate the listening resource.

May be @talex5 can do an other view about `lwt` stuffs and the `switch_off` field but it seems fine for me :+1:.

hannesm:
strangely somehow the CI (analysis phase), even after "rebuild", does not seem to know anything about mirage-protocols 6.0.0 (merged yesterday https://github.com/ocaml/opam-repository/commit/df836af003c106046b95469da90c464c6ddeaed2) -- any insights how/when/why this happens?

talex5:
We recently added signature checks for GitHub events. Possibly the opam-repository configuration needs updating with a secret. I'll see if I can find someone with access to that...

talex5:
OK, should be building now - thanks to @dra27!

hannesm:
CI succeeded, apart from ppc64 which timed out.

PR #465, opened 2021-12-15
stack-unix: use custom switched_off values for udp & tcp socket stacks

### LINKED ISSUE #464: `tcpip.stack-unix` has a bug about `accept` when we initiate a TCP connection
This is related to the recent update on #457 when, if we only initiate a TCP connection and we don't use `Tcpip_socket.V4V6.connect`, the `switched_off` fiber is never updated into the TCP stack and we always cancel the `listen` fiber here:
https://github.com/mirage/mirage-tcpip/blob/3ab30ab7b43dede75abf7b37838e051e0ddbb23a/src/stack-unix/tcpv4v6_socket.ml#L154

A solution should be to set by default a fiber on the `switched_off` value which never finish:
```ocaml
let never, _ = Lwt.wait () in
Lwt.return {interface; active_connections = []; listen_sockets = Hashtbl.create 7; switched_off = never}
```

/cc @hannesm, @talex5

hannesm:
this improves #457 and fixes #464

### DIFF
--- a/src/stack-unix/tcpip_stack_socket.ml
+++ b/src/stack-unix/tcpip_stack_socket.ml
@@ -46,8 +46,6 @@ module V4 = struct
   let connect udpv4 tcpv4 =
     Log.info (fun f -> f "IPv4 socket stack: connect");
     let switched_off, stop = Lwt.wait () in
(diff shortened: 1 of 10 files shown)


hannesm:
superseeded by #466

PR #466, opened 2021-12-15
Fix cancelation of UNIX sockets and let a chance to run the listen loop if we craft our TCP stack xor UDP stack on our own way

dinosaure:
This is a probable fix about cancellation of sockets and a fix for #464. I would like a double check for an lwt expert (@talex5 ?) and @hannesm :+1:!


hannesm:
ok, I suspect that for udp{v4,v6,v4v6}_socket.ml the same should be applied.

dinosaure:
I'm not sure if @talex5 is available to look on it but the PR seems ok to merge.

dinosaure:
Let's merge, I'm confident with this fix.

hannesm:
ok, cool. should I cut a release with this fix? or do you want to cut a release?

dinosaure:
I can cut the release by myself 👍 .

PR #472, opened 2022-01-25
freestanding/Makefile: remove 'which', 'command -v' is sufficient

hannesm:
//cc @MisterDA @TheLortex @dinosaure


dinosaure:
Ah thanks, faster than me, waiting the CI and we will merge it.

PR #482, opened 2022-04-09
ndpv6: Set RS opt header

reynir:


hannesm:
checked with RFC 4861 this looks very reasonable. Thanks!

PR #484, opened 2022-04-14
Namespace tcp log sources

reynir:
This adds a `tcp.` prefix to all tcp log sources. This makes it more clear where the log source is coming from (e.g. `state` could be many things).

An additional change I consider making is to make `tcp.Wire` all lower-case as in the other log sources.

### DIFF
--- a/src/tcp/wire.ml
+++ b/src/tcp/wire.ml
@@ -15,7 +15,7 @@
  *)
 open Lwt.Infix
 
-let src = Logs.Src.create "Wire" ~doc:"Mirage TCP Wire module"
+let src = Logs.Src.create "tcp.Wire" ~doc:"Mirage TCP Wire module"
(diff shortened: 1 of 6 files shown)


src/tcp/wire.ml
```diff
@@ -15,7 +15,7 @@
  *)
 open Lwt.Infix
 
-let src = Logs.Src.create "Wire" ~doc:"Mirage TCP Wire module"
+let src = Logs.Src.create "tcp.Wire" ~doc:"Mirage TCP Wire module"
```
hannesm, line 18:
```suggestion
let src = Logs.Src.create "tcp.wire" ~doc:"Mirage TCP Wire module"
```

hannesm:
> An additional change I consider making is to make tcp.Wire all lower-case as in the other log sources.

That sounds great! I did this directly here.

Issue #488, opened 2022-06-10
Memory issues in TCP


hannesm:
I tried to investigate some memory leaks of https://github.com/roburio/tlstunnel, and put some printf debugging into mirage-tcpip (see this commit https://github.com/hannesm/mirage-tcpip/commit/f358e782abd1214ef30dc5e60dc43945804b09d6) which prints some statistics about buffers that are present. Here are the numbers:
```
tcp.pcb-stats removing pcb from connection tables: [channels=460 (RX 20 TX 12993900 URX 5735 UTX 54732 TOTAL 13054387) listens=0 (RX 0 TX 0 URX 0 UTX 0 TOTAL 0) connects=0]
tcp.pcb-stats process-synack: [channels=215 (RX 0 TX 0 URX 45842733 UTX 0 TOTAL 45842733) listens=0 (RX 0 TX 0 URX 0 UTX 0 TOTAL 0) connects=1]
```

There are two TCP/IP stacks in the unikernel (one public, which is the top line, with listening sockets; one private where client connections are initiated to the backend servicces).

Now, the observation is: there are quite some channels open (460 / 215) -- which I'd have not expected to see so many. Furthermore there is quite some data in the TX queue and the URX (awaiting an application to call read).

I may be doing something completely stupid in tlstunnel, but I thoought that the two reader/writer functions should be fine. That's even from a branch including more aggressive `close` https://github.com/roburio/tlstunnel/commit/0b985abc2c04badd6fd08b2713305e4f537a017d.

I'm wondering what is up with TCP: are there some Lwt_mvar.t that are stuck? Is a `close` not cleaning up everything? Is there a simple example on how to properly write a TCP server and client in a way that they always free up all resources after a while? I also had the impression that eventually the window size is transmitted via TCP, but not respected in User_buffer (i.e. when fresh segments come in, they are not discarded, but instead are acked).

I don't have singled out TCP flows that cause such stuck states, but eventually someone else has some further ideas (though, looking at #486 and #470 it doesn't look like there's much activity here).

PR #489, opened 2022-06-30
Fixing memory leaks when a simultaneous close is happening

TheLortex:
When the client and the server want to close a connection at the same time, they both send a `FIN`. 

I have identified situations where `mirage-tcpip` might keep the pcb forever, stuck in closing states, causing a memory leak.
This might be linked to https://github.com/mirage/mirage-tcpip/issues/488.

This PR adds two tests exposing the behavior. 

## Test 1: `close_ack_scenario`

Both client and server sends a FIN, then they ACK each other's FIN. The state transition for the server is:
- `Established  - Send_fin(1) -> Fin_wait_1(1)`: server FIN is sent
- `Fin_wait_1(1)  - Recv_ack(1) -> Fin_wait_1(1)`: previous packet is acked
- `Fin_wait_1(1)  - Recv_fin -> Closing(1)`: client FIN is received
- `Closing(1)  - Recv_ack(2) -> Time_wait`: server FIN is acked 

The problem is that the connection state would stay in `Time_wait` forever.
This is fixed by commit [Add timewait timeout for Closing -> Time_wait transition](https://github.com/mirage/mirage-tcpip/commit/bb6e85d29e048a2ecf730e5737e231e4723db596).

## Test 2: `close_reset_scenario`

Now, we assume the client doesn't ackowledge the FIN. The connection is then in a `Closing(_)` state. 

I have encountered a situation where it would send a `RST` instead with a valid but not expected sequence number. In that case, `check_valid_segment` returns `ChallengeAck` and writes into `q.rx_data` but because we are in a closing state that mailbox variable is not read anymore (as in flow.ml `Rx.thread` would have returned). So the challenge ack is not sent and the connection stay in this zombie state.

The fix for this is to use the `send_ack` mailbox variable that can be used to send an empty ack even if the rx thread is stopped:
[enable challenge ack even after FIN](https://github.com/mirage/mirage-tcpip/commit/b9b88d956279bc8b31732cf063f2541fd21c427d)

## Untested: What if the client stops communicating ? 

In this `Closing(_)` state, we are waiting for the client to ACK the server's FIN. If nothing happens, the retransmission mechanism should re-send the FIN. Instead, the retransmission timer is currently stopped if the connection is in the `Closing(_)` state 

[Add retransmit timer when in a Closing state](https://github.com/mirage/mirage-tcpip/commit/c088044e8e70347ffb858ef93f3ea8264606aafe) 

 # Disclaimer

TCP is hard ! I would be very happy if someone can review these changes and make sure nothing is broken with these patches.


haesbaert:
This is a great find ! :1st_place_medal: 
I'm back from :palm_tree: next week and will have a look

ansiwen:
Thanks folks for looking into this and thanks a lot @TheLortex for the great work so far!

Some context: This work resulted from a memory leak that we only can observe with an unikernel running as a Muen subject (=VM). As a kvm instance I can't reproduce the issue. Also it only shows up when the requests (from curl in this case) are directly done on the network interface connected to the server hardware Muen is running on. If the request comes from another routed network (from a container for example) it also doesn't show up. The sad news is: If I didn't do something wrong, this fix still doesn't fix our issue. There seems to be more luring. So it would be great if you folks could look for something more that might be leaking.

A detail that might be relevant, is that the clock provided by the Muen kernel has a reduced resolution for security reasons. Not sure this is really relevant, but just to keep it in mind.

Here some info from our internal ticket:

I created pcap files and debug logs for both cases. As you can see, the debug logs show, that in the leaking case, the pcb entry is not removed:
Bildschirmfoto 2022-06-27 um 16 11 52

At the same time the pcap traces show that in the leaking case only one RST package is sent, in the non-leaking case there are four, but I'm not sure if this is relevant:

**without leak:**
trace-no-leak
**with leak:**
trace-leak

Here the pcap files with 1000 requests for both cases (with and without leak):
[leak-info-1000-without-leak.pcap.zip](https://github.com/mirage/mirage-tcpip/files/9032578/leak-info-1000-without-leak.pcap.zip)
[leak-info-1000-with-leak.pcap.zip](https://github.com/mirage/mirage-tcpip/files/9032579/leak-info-1000-with-leak.pcap.zip)

src/tcp/state.ml
```diff
@@ -151,7 +151,13 @@ module Make(Time:Mirage_time.S) = struct
       | Fin_wait_2 _, Recv_fin ->
         Lwt.async (fun () -> timewait t time_wait_time);
         Time_wait
-      | Closing a, Recv_ack b -> if diffone b a then Time_wait else Closing a
+      | Closing a, Recv_ack b ->
```
haesbaert, line 154: white-space eol, most of this lines have a blank white-space added in the end

src/tcp/state.ml
```diff
@@ -151,7 +151,13 @@ module Make(Time:Mirage_time.S) = struct
       | Fin_wait_2 _, Recv_fin ->
         Lwt.async (fun () -> timewait t time_wait_time);
         Time_wait
-      | Closing a, Recv_ack b -> if diffone b a then Time_wait else Closing a
+      | Closing a, Recv_ack b -> 
+        if diffone b a then 
+          begin 
+            Lwt.async (fun () -> timewait t time_wait_time);
```
haesbaert, line 157: I agree the fix is correct.
It seems to always be an error to just transition into Time_wait without arming the timer, maybe we should change `timewait` to actually perform all of it and already return `Time_wait`

src/tcp/flow.mli
```diff
@@ -20,4 +20,10 @@ module Make (IP:Tcpip.Ip.S)
             (R:Mirage_random.S) : sig
   include Tcpip.Tcp.S with type ipaddr = IP.ipaddr
   val connect : IP.t -> t Lwt.t
+
+  (**/**)
+  (* the number of open connections *)
+  val internal_n_channels : t -> int
```
haesbaert, line 26: A bit of bike-shedding but why not just `num_channels` ? or even `num_open_channels`

src/tcp/segment.ml
```diff
@@ -135,10 +136,7 @@ module Rx(Time:Mirage_time.S) = struct
 
   let send_challenge_ack q =
     (* TODO:  rfc5961 ACK Throttling *)
-    (* Is this the correct way trigger an ack? *)
-    if Lwt_mvar.is_empty q.rx_data
-      then Lwt_mvar.put q.rx_data (Some [], Some Sequence.zero)
-      else Lwt.return_unit
+    Lwt_mvar.put q.send_ack Sequence.zero
```
haesbaert, line 139: I understand you need the new mvar since the rx is dead, but why did you remove the `is_empty` dance ?
I'm more comfortable with the way you did but I'm pondering what made you choose this

haesbaert (commented):
Here is a partial review with some questions, I still need to understand the ACK dance, but I'll do it tomorrow.

src/tcp/segment.ml
TheLortex: That's a mistake from my side, it's probably better to keep the `is_empty` check.

TheLortex:
Yes exactly. RST with Seq=804 is supposed to trigger a challenge ACK, as it is not the next expected Seq number, as the previous FIN also uses Seq=804. That's why only RST with Seq=805 is handled.

hannesm:
I have not looked into the details of this PR, but @ansiwen reported there's still a leak: is the pcap the same as before this patch, or is there a difference?

ansiwen:
Because mirage ignored the pin in `config.ml` I didn't actually try the patched version before. I pinned now directly in opam and then it finally worked!

Here is how the debug log looks like now (right) compared to before (left):
Bildschirmfoto 2022-07-19 um 17 15 46

The network trace looks like this now:
Bildschirmfoto 2022-07-19 um 16 24 29

[leak-info-1000-baremetal-with-fix4.pcap.zip](https://github.com/mirage/mirage-tcpip/files/9138974/leak-info-1000-baremetal-with-fix4.pcap.zip)

src/tcp/segment.ml
```diff
@@ -135,9 +136,8 @@ module Rx(Time:Mirage_time.S) = struct
 
   let send_challenge_ack q =
     (* TODO:  rfc5961 ACK Throttling *)
-    (* Is this the correct way trigger an ack? *)
-    if Lwt_mvar.is_empty q.rx_data
-      then Lwt_mvar.put q.rx_data (Some [], Some Sequence.zero)
+    if Lwt_mvar.is_empty q.send_ack
+      then Lwt_mvar.put q.send_ack Sequence.zero
       else Lwt.return_unit
```
haesbaert, line 141: If I read this correctly, the original `Ack.t` stored in `Flow.flow` (aka `Flow.pcb`) is not reachable through `Segment.Rx.t`, which is a shame, maybe `ack` should have been stored deeper.

So it begs the question why is your `send_ack` not a `Ack.t`, so instead of pushing something directly on a mailbox you could do the same `Ack.pushack` dance that the `Rx.thread` do, something like:

```ocaml
  let send_challenge_ack q =
    (* TODO:  rfc5961 ACK Throttling *)
    Ack.pushack q.ack Sequence.zero
```

TheLortex: I am not familiar with the `Ack` module. But indeed it looks like it's doing similar ! 

I have added a commit that uses the `pushack` function as suggested. 
Now the `Segment.Rx` module is functorized against the `Ack.M` signature.

haesbaert (approved):
That looks good to me, I just hope my suggestion didn't break the fix somehow :).
Thanks for the work, this is a great find !

haesbaert:
Can we get another OK for this  ?

TheLortex:
@hannesm or @balrajsingh would one you be available to review the patch ? We're now confident that it fixes Nitrokey's issue.

hannesm:
Code looks fine (though I'm still not an expert at this TCP/IP stack), @ansiwen reports it solves the issue. Would you mind to write an entry for CHANGES and cut a release @TheLortex (I can as well do the release if you like).

hannesm:
Question for @ansiwen: did you test [1813264](https://github.com/mirage/mirage-tcpip/pull/489/commits/1813264ba564136b0400f7f4a66034ea05fa24e7) or which commit?

TheLortex:
> Would you mind to write an entry for CHANGES and cut a release @TheLortex (I can as well do the release if you like).

Great, I'll take care of that.

@ansiwen tested on 1bd6270b2b4a3f3974a0402b44aa61a984f2d8ab, and normally the additional commit 1813264ba564136b0400f7f4a66034ea05fa24e7 does not change the behavior of the stack

balrajsingh:
I mainly looked to see if the `send_challenge_ack` change breaks anything else - this is the central change as I understand it, please let me know if not so.

I think that the fix is good.  It is specific to the challenge ACK situation only, when no data needs to be pushed up to the application.  Your test is also convincing.  I too agree that the change can be merged.

And @TheLortex, yes TCP is hard.  So glad that you are looking at it in such detail.  Many thanks!

ansiwen:
> @ansiwen tested on 1bd6270b2b4a3f3974a0402b44aa61a984f2d8ab, and normally the additional commit 1813264ba564136b0400f7f4a66034ea05fa24e7 does not change the behavior of the stack

No, actually I tested still on b9b88d956279bc8b31732cf063f2541fd21c427d, I will test tomorrow again with the latest commit.

TheLortex:
I have added a changelog entry and I am ready to cut a patch release today when @ansiwen confirms that the latest commit still works !

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+### v7.1.2 (2022-07-27)
+
+* TCP: fix memory leaks on connection close (#489 @TheLortex)
```
hannesm, line 3: In the issue comment you had quite a bit more information which scenarios were dealt with, it may be worth to be a bit more wordy in the changes entry as well...

TheLortex: I have force-pushed with aac0e02f3b2e3da016f6f7cfb9790c4949be8456 containing more information about the various patches

ansiwen:
I can confirm that 1813264ba564136b0400f7f4a66034ea05fa24e7 still fixes the issue.

dinosaure:
Let's merge and cut a release so! Thanks all for your work.

PR #490, opened 2022-07-26
[WIP] Initial attempt at rfc8106

hannesm:
Signed-off-by: Jonathan Davies <jpds@protonmail.com>

### DIFF
--- a/src/ipv6/ipv6_wire.ml
+++ b/src/ipv6/ipv6_wire.ml
@@ -104,6 +104,15 @@ let get_opt_prefix_autonomous buf =
   get_opt_prefix_reserved1 buf land 0x40 <> 0
 
 [%%cstruct
+type opt_rdnss_header = {
+    ty:                 uint8_t;
+    len:                uint8_t;
+    reserved:           uint16_t;
+    rdnss_lifetime:     uint32_t;
+  } [@@big_endian]
+]
+
+[%%cstruct
 type opt = {
     ty:  uint8_t;
     len: uint8_t;
--- a/src/ipv6/ndpv6.ml
+++ b/src/ipv6/ndpv6.ml
@@ -34,6 +34,9 @@ References:
 
 - Multicast Listener Discovery Version 2 (MLDv2) for IPv6
   http://tools.ietf.org/html/rfc3810
+
+- IPv6 Router Advertisement Options for DNS Configuration
+  https://tools.ietf.org/html/rfc8106
 *)
 
 let src = Logs.Src.create "ndpc6" ~doc:"Mirage IPv6 discovery"
@@ -255,12 +258,17 @@ type pfx =
     pfx_preferred_lifetime : time option;
     pfx_prefix : Ipaddr.Prefix.t }
 
+type rdnss = 
+  { rdnss_lifetime: time option;
+    rdnss_addresses: Ipaddr.t list }
+
 type ra =
   { ra_cur_hop_limit : int;
     ra_router_lifetime : time;
     ra_reachable_time : time option;
     ra_retrans_timer : time option;
     ra_slla : Macaddr.t option;
+    ra_rdnss : rdnss list;
     ra_prefix : pfx list }
 
 type na =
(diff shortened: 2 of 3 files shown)


src/ipv6/ipv6_wire.ml
```diff
@@ -103,6 +103,15 @@ let get_opt_prefix_on_link buf =
 let get_opt_prefix_autonomous buf =
   get_opt_prefix_reserved1 buf land 0x40 <> 0
 
+[%%cstruct
+type opt_rdnss = {
+    ty:                 uint8_t;
+    len:                uint8_t;
```
hannesm, line 109: hereafter we need a `reserved : uint16_t` (since there's some padding, and the lifetime starts at offset 4)

src/ipv6/ipv6_wire.ml
```diff
@@ -103,6 +103,15 @@ let get_opt_prefix_on_link buf =
...
+    rdnss_lifetime:     uint32_t;
+    rdnss_addresses:    uint8_t [@len 16];
```
hannesm, line 111: keep in mind that the addresses are any number of addresses (actually "len - 1" ones) -- but I can't see a way in ppx_cstruct to define "now there's a list of 16 byte values, in the number of len - 1". so it is fine to leave this as is (maybe add a comment), but the actualy decoding and encoding part needs to be aware of that.

src/ipv6/ndpv6.ml
```diff
@@ -270,6 +274,10 @@ type na =
     na_target : Ipaddr.t;
     na_tlla : Macaddr.t option }
 
+type rdnss = 
+  { rdnss_lifetime: time option;
+    rdnss_addresses: Ipaddr.t }
```
hannesm, line 279:
```suggestion
    rdnss_addresses: Ipaddr.V6.t list }
```

src/ipv6/ndpv6.ml
```diff
@@ -777,6 +840,22 @@ module Parser = struct
...
         in
         PREFIX pfx :: parse_options1 opts
+      | 25, 3 ->
+        let rdnss_lifetime =
+          let n = Ipv6_wire.get_opt_rdnss_rdnss_lifetime opt in
+          match n with
+          | 0l -> None
+          | n -> Some (Int64.of_int32 n)
+        in
+        let rdnss_addresses =
```
hannesm, line 850: here the decoding happens, and `len - 1` needs to be looked at to figure out how many ip addresses to parse. all of them are IPv6 addresses, that's why I'd use Ipaddr.V6.t.

src/ipv6/ndpv6.ml
```diff
@@ -665,6 +673,60 @@ module NeighborCache = struct
...
+  type t =
+    (Ipaddr.t * time) list
+
+  let empty =
+    []
+
+  let to_list rdnssl =
+    List.map fst rdnssl
+
+  let add rdnssl ~now ?(lifetime = Duration.of_year 1) ip =
```
hannesm, line 687: From RFC 8106:
The value of Lifetime SHOULD by default be at least 3 * MaxRtrAdvInterval

Can we use that as default, and not 1 year :)

src/ipv6/ipv6_wire.ml
hannesm: so maybe just skip the `rdnss_addresses` entirely here (and have the `type opt_rdnss_header`)?

src/ipv6/ndpv6.ml
hannesm: If we remove from the `type opt_rdnss` the `rdnss_addresses`, we could:
```OCaml
let decode_ns off = Ipaddr_cstruct.V6.of_cstruct_exn (Cstruct.shift opt off) in
let rec collect_ns acc = function
   | 0 -> acc
   | n ->
    let ns = decode_ns (Ipv6_wire.sizeof_opt_rdnss + n * 16) in
    collect_ns (ns :: acc) (n - 1)
in
let rdnss_addresses = collect_ns [] (Ipv6_wire.get_opt_rdnss_len opt - 1) in
```

src/ipv6/ndpv6.ml
jpds: I'm not sure how to access MaxRtrAdvInterval here as it's something that's set by the router.

src/ipv6/ndpv6.ml
jpds: `Ipaddr` is already aliased(?) to `Ipaddr.V6` in this file.

src/ipv6/ndpv6.ml
reynir: RFC 4861 says about MaxRtrAdvInterval:

> MUST be no less than 4 seconds and no greater than 1800 seconds.

So perhaps 1½ hours is reasonable.

Reading the code and some parts of RFC 8106 it seems to me that the lifetime is only `None` when it's zero in the packet, and in the rfc it says

> A value  of zero means that the RDNSS addresses MUST no longer be used

So maybe we should either promote `lifetime` to a regular labeled argument and optionally assert lifetime is greater than zero, or we should not have a default value and do nothing if `lifetime` is `None`.

I wonder if something similar applies to `RouterList.add` where there's a FIXME comment about the default lifetime...

reynir: ... or perhaps instead of doing nothing we should actively remove that IP address from the list.

tmcgilchrist:
Is there interest in continuing work to support RFC-8106? @hannesm @jpds

dinosaure:
> Is there interest in continuing work to support RFC-8106? @hannesm @jpds

I think it would be more interesting to move this work over to [`mnet`](https://github.com/robur-coop/mnet).

PR #494, opened 2022-10-04
remove V4-only and V6-only stack (as done in mirage 4.3.0)

hannesm:


hannesm:
fine to be merged, as a separate line of work we can address 3ccba30b9d0969740b6c7d8b47ea76d9e7f14ae5 and remove code duplication:
```
This duplicates the iperf test for ipv6. As the V4 stack signatures
contain a reference to the IP version (e.g. listen_tcpv4 etc) it was
difficult to reuse the test without major changes. We should clean this
up when we unify the signatures to avoid duplicated tests for IPv4/IPv6.
```

any objections in this PR?

PR #495, opened 2022-10-06
TCP: Add ID for PCB to better track connections

TheLortex:
As I'm investigating a connection leak, I found it useful to annotate state transitions with an unique ID per connection.

### DIFF
--- a/src/tcp/flow.ml
+++ b/src/tcp/flow.ml
@@ -318,7 +318,7 @@ struct
 
   let emitted_keepalive_warning = ref false
 
-  let pcb_id = ref 0
+  let pcb_id = ref (-1)
 
   let new_pcb t params id keepalive =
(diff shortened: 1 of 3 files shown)


src/tcp/flow.ml
```diff
@@ -354,7 +356,10 @@ struct
     let tx_wnd_update = MProf.Trace.named_mvar_empty "tx_wnd_update" in
     (* Set up transmit and receive queues *)
     let on_close () = clearpcb t id tx_isn in
-    let state = State.t ~on_close in
+    let state =
+      incr pcb_id;
+      State.t ~id:!pcb_id ~on_close
```
haesbaert, line 361: start on zero please :)

hannesm:
to me this looks good. I agree with @haesbaert, let's start at 0.

Issue #501, opened 2023-02-09
`S.UDP.listen` blocks differently on `unix` vs `direct` stack 


rand00:
I found that when giving a callback to `S.UDP.listen`, the `unix` stack blocks on the callback before receving next UDP datagram - but the `direct` stack doesn't. 

My usecase has been to implement TCP-like semantics on top of UDP in the [`conntest`](https://github.com/rand00/conntest/blob/526e2bf6b4b3d3691601dc078dfef9c0e0df9576/lib/conntest.ml#L444) unikernel. As the lifetime of the callback in this case is as long as the 'UDP connection', then the callback need to be run async and potentially saved elsewhere for managing the state of the thread. 

This is fine for me in this case (when I know of the semantics), but I was initially confused that the UDP callback wasn't run right away when a datagram was received - this could become a more subtle problem to find for some users, if their callback takes a bit of time but they expect it to be run instantly on each datagram. Off the top of my head this can influence:
* performance of code using UDP
* potential loss of datagrams if some underlying buffer gets filled before next datagram is read by client (I'm unaware of the actual underlying semantics)

Otoh. when callbacks are run concurrently, it can lead to a need for the client to synchronize between the callbacks if they mutate anything.

In any case, documentation of the semantics would be nice, and having the same behaviour across stacks.

rand00:
.. and if callbacks are concurrent, it can lead to memory-leak, as packets can get received faster than callbacks finish up.

hannesm:
In  this callback https://github.com/mirage/mirage-tcpip/blob/main/src/stack-unix/udpv4v6_socket.ml#L194 in the Unix stack is run synchronously with Lwt_cstruct.recvfrom (so there won't be multiple callbacks executed concurrently) -- while in the Mirage stack due to every mirage-net-* implementation calling "Lwt.async" for the callback of each incoming packet, the MirageOS UDP callback is asynchronous (i.e. there can be multiple callbacks concurrently).

I still don't quite understand what the desired semantics should be -- maybe the callback should not involve "io / Lwt.t" at all? Maybe it should be all sync? My intuition is that the "Lwt.async" for each incoming packet is not optimal from the performance and scheduling perspective.

rand00:
But I guess that user wants to run stuff within the callback in the Lwt monad? How else would one communicate nicely with the rest of the Lwt code? 

[...]

PR #502, opened 2023-02-09
Unix: copy buffer before passing off to client

reynir:
Originally reported by @rand00.

### DIFF
--- a/src/stack-unix/udpv4v6_socket.ml
+++ b/src/stack-unix/udpv4v6_socket.ml
@@ -191,6 +191,8 @@ let listen t ~port callback =
                          in
                          let dst = Ipaddr.(V6 V6.unspecified) in (* TODO *)
                          let buf =
+                           (* Use Cstruct.sub_copy once it exists in a
+                              reasonably mature cstruct release *)
                            let b = Cstruct.create_unsafe len in
                            Cstruct.blit buf 0 b 0 len;
                            b
--- a/src/stack-unix/udpv4v6_socket.ml
+++ b/src/stack-unix/udpv4v6_socket.ml
@@ -181,7 +181,6 @@ let listen t ~port callback =
                   if not (Lwt.is_sleeping t.switched_off) then raise Lwt.Canceled ;
                   Lwt.catch (fun () ->
                       Lwt_cstruct.recvfrom fd buf [] >>= fun (len, sa) ->
-                      let buf = Cstruct.sub buf 0 len in
                       (match sa with
                        | Lwt_unix.ADDR_INET (addr, src_port) ->
                          let src = Ipaddr_unix.of_inet_addr addr in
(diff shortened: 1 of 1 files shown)


src/stack-unix/udpv4v6_socket.ml
```diff
@@ -191,6 +190,11 @@ let listen t ~port callback =
                            | Some v4 -> Ipaddr.V4 v4
                          in
                          let dst = Ipaddr.(V6 V6.unspecified) in (* TODO *)
+                         let buf =
+                           let b = Cstruct.create_unsafe len in
+                           Cstruct.blit buf 0 b 0 len;
+                           b
+                         in
                          callback ~src ~dst ~src_port buf
```
hannesm, line 198: What I understood from the report #501 is that here this callback in the Unix stack is run synchronously with Lwt_cstruct.recvfrom (so there won't be multiple callbacks executed concurrently) -- while in the Mirage stack due to every mirage-net-* implementation calling "Lwt.async" for the callback of each incoming packet, the MirageOS UDP callback is asynchronous (i.e. there can be multiple callbacks concurrently).

I still don't quite understand what the desired semantics should be -- maybe the callback should not involve "io / Lwt.t" at all? Maybe it should be all sync? My intuition is that the "Lwt.async" for each incoming packet is not optimal from the performance and scheduling perspective.

reynir: I think this PR is orthogonal to #501. @rand00 also reported a buffer ownership bug on IRC (but no issue AFAICT).

hannesm: ah, ok :)

src/stack-unix/udpv4v6_socket.ml
```diff
@@ -191,6 +190,11 @@ let listen t ~port callback =
                            | Some v4 -> Ipaddr.V4 v4
                          in
                          let dst = Ipaddr.(V6 V6.unspecified) in (* TODO *)
+                         let buf =
+                           let b = Cstruct.create_unsafe len in
+                           Cstruct.blit buf 0 b 0 len;
+                           b
+                         in
```
hannesm, line 197: maybe use Cstruct.copy?

reynir: Cstruct.copy is Cstruct.t -> int -> int -> string, that is, `Cstruct.copy t off len` is the same as `Cstruct.to_string t ~off ~len` :/ I didn't see a function to copy a cstruct (even though Cstruct.copy would be an obvious name for such a function)

reynir: Hi @palainp, indeed, `Cstruct.sub_copy` does what I want, but it only exists in unreleased cstruct. I can put a comment in the code to use `Cstruct.sub_copy` in the future.

palainp: Oh sorry, didn't checked the last release date :(

dinosaure: I will cut a release of `cstruct` as soon as I can but some improvements are not merged yet.

dinosaure (approved):

PR #503, opened 2023-02-09
Use Lwt.dont_wait instead of async (catch ...)

reynir:
Just a minor nit in stack-unix. It could use some eyes, it seems :/


src/stack-unix/tcpv4v6_socket.ml
```diff
@@ -161,13 +161,10 @@ let listen t ~port ?keepalive callback =
...
-                  (fun () ->
-                     Lwt.catch
-                       (fun () -> callback afd)
-                       (fun exn ->
-                          Log.warn (fun m -> m "error %s in callback" (Printexc.to_string exn)) ;
-                          close afd));
+                Lwt.dont_wait (fun () -> callback afd)
+                  (fun exn ->
+                     Log.warn (fun m -> m "error %s in callback" (Printexc.to_string exn)) ;
+                     close afd);
```
reynir, line 167: I thought the signature for `Lwt.dont_wait` was different. To fix it we can do:
```suggestion
                     Lwt.async (fun () -> close afd));
```

PR #505, opened 2023-03-09
remove ppx from mirage-tcpip

hannesm:
The rationale is that ppx_cstruct is more trouble than worth it:
- take a look at the intricate hoops in ipv6_wire how to retrieve the ipv6 address from a `[%%cstruct ]`
- take a look at the dependency chain: requires sexplib, which requires base, which takes several minutes to compile

Also, the tooling (merlin go to source) of ppx, and the readability is very bad (I've seen several beginners who wanted to dig into ipv6/ipv4 stack, and had a hard time to figure out where these magic functions are coming from).

Please raise your voice if you're concerned, or explain the reasoning why we should keep the ppx_cstruct around.

### DIFF
--- a/src/icmp/dune
+++ b/src/icmp/dune
@@ -4,6 +4,4 @@
  (instrumentation
   (backend bisect_ppx))
  (libraries logs tcpip ipaddr tcpip.checksum)
- (preprocess
-  (pps ppx_cstruct))
  (wrapped false))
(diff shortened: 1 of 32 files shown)


hannesm:
I revised the diff to reduce the API change (in the `_wire` modules,  only the accessors now have dropped the name of the type (e.g. `get_icmpv4_ty` is now `get_ty`). These wire modules are intended for private use anyways. This PR is fine to review and merge.

hannesm:
I did some measurements, FWIW of interest is trusted code base and compilation time:

|    | tcpip 7.1.2 | this PR |
|---|---|---|
| dependencies | 45 | 35 |
| build time | 227.979u 67.019s | 123.088u 42.661s |
| opam switch size | 225 MB | 140 MB |

The build time was from an empty switch (OCaml 4.14.1) installing tcpip and its dependencies. The switch size was evaluated by `du -hs ~/.opam/<switch>/lib`.

The result is roughly a factor of 2 in build time and also in trusted code base. WDYT?

palainp:
This seems great to me! Thanks a lot for this work!

avsm:
Impressive build time improvement as well. Nice work!

PR #511, opened 2023-12-16
add TCP.S.src

hannesm:
Before, we weren't able to uniquely identify a TCP flow (esp. for client connection to the same destionation host and port).

With this change, we are finally able to print a proper quad (src_ip, src_port, dst_ip, dst_port). Adding this to the interface is trivial and pretty useful.


palainp (approved):
Thanks, this is good for me!

reynir (approved):

PR #512, opened 2023-12-21
Implement shutdown, as required by mirage-flow 4.0.0

hannesm:
Would be great to have a review on this.

## close
From my understanding, in the UNIX sockets API the SO_LINGER is relevant (to specify how long `close()` may block) - since we don't have SO_LINGER, let's just care about the normal case (text paraphrased from https://www.cl.cam.ac.uk/~pes20/Netsem/alldoc.pdf):
- close sends any remaining data and gracefully closes the connection
- if close is called on a pre-established state: socket is closed and removed
- if on a listening socket: close and remove that socket, and aborts each of the connections on the socket's pending and completed connection queues.

If I understand, our `flow` is never a listening socket. Our `unlisten` eventually should care about the last point.

## shutdown
Depending on `` `read | `write | `read_write ``, the specific half is shut down:
- read: shutting down the read-half empties the socket’s receive queue, but data will still be delivered to it and subsequent recv() calls will return data.
- write: Shutting down the write-half of a TCP connection causes the remaining data in the socket’s send queue to be sent and then TCP’s connection termination to occur.

## TL;DR

There's some discrepancy from my intuition (and the man page shutdown(2) SHUT_RD and the model: according to the man page) "Further receives will be disallowed". In this PR, I push None to the Lwt_dllist -- with the hope that this leads to the expected behaviour (those who have knowledge about the internals should comment whether there's something else that needs to be done).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,9 @@ jobs:
       - name: Use OCaml ${{ matrix.ocaml-compiler }}
         uses: ocaml/setup-ocaml@v2
         with:
+          opam-repositories: |
+            opam-repository-mingw: https://github.com/ocaml-opam/opam-repository-mingw.git#sunset
+            default: https://github.com/ocaml/opam-repository.git
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
 
       - run: opam install . --deps-only --with-test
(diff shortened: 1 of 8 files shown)


djs55 (approved):
LGTM. I'm not sure exactly how the `SHUTDOWN_RECEIVE` is supposed to behave but I've never found a use for it. The `SHUTDOWN_SEND` to perform a half-close is really useful though and this looks good.

hannesm:
Thanks @djs55. I'll go ahead with merge and release once CI is green.

PR #516, opened 2024-05-09
IP: add configured_ips : t -> prefix list

hannesm:


hannesm:
This PR has been updated, `get_ips` is now deprecated, and in this package not used anymore.

hannesm:
This requires the ipaddr package as proposed in https://github.com/ocaml/opam-repository/pull/25919 -- so we'll need to wait for that being merged before the CI in here will succeed.

hannesm:
tests were failing, there's a discrepancy in ipv6 with having both prefix and address list.

I changed the address list to contain Ipaddr.Prefix.t now -- this means both the IP address and the netmask is stored in there. The tests pass. IMHO good to merge. We'll need to do an IPv6 deep dive at some point in any case.

PR #520, opened 2024-07-22
Fix compilation of tests

samoht:


samoht:
A better fix would be to use `mirage-crypto.1.0.0` and merge #518 that started to add support for it.

PR #522, opened 2024-08-05
update unikernel to recent mirage

hannesm:


hannesm:
The issue is that tcpip is pinned to `tcpip.dev`, while mirage requires it to be `>= 7.0.0 & < 9.0.0`... Is there any support in setup-ocaml to pin packages to a specific version @smorimoto?

smorimoto:
In this case, just pass `false` to the `opam-pin` input is enough!

hannesm:
> In this case, just pass `false` to the `opam-pin` input is enough!

Thanks for your quick reply, unfortunately I don't quite understand where in the main.yml a `false` needs to be passed. Could you provide me with some hint?

reynir:
I found in https://github.com/ocaml/setup-ocaml/?tab=readme-ov-file#inputs the `ocaml-compiler` option next to the `opam-pin` option and did pattern matching (qualified guess) to figure out where to add it. Let's see if it works. I hope I am not stepping on anyone's toes by pushing this commit.

hannesm:
Thanks a lot, merging... CI is green :)

PR #523, opened 2024-12-06
dune: enable generate_opam_files option

smorimoto:


hannesm:
what is the motivation for this PR? I appreciate the CI fixes, which could be a separate PR.

But the generate_opam_file, I'm rather hesitant unless there's a clear motivation. Now, you don't even provide a description, so I have to ask.

smorimoto:
Today I had to debug this package, and since I've been using dune pkg lately, I tried to bootstrap the project using it, but it complained so I moved it to the dune-based workflow as much as possible.

If it's OK, you can merge it, or if you're not interested, you can close! Then, of course I can separate the CI fix into another PR.

PR #528, opened 2025-04-11
unix stack, UDP: if 0 is returned from recvfrom, do not further read data

hannesm:
0 is an indication of eof.

Took me some time to figure out why a UDP socket application takes 100% CPU... but that was a very busy loop here.


hannesm:
anyone happy with this, or against it?

reynir (approved):
Looks good to me.

PR #529, opened 2025-06-04
Readd ipv4 get_tos and set_tos

patricoferris:
Seems this was lost in the transition away from ppx_cstruct.


hannesm:
I'm curious, why do you need that? A specific protocol and use case, or just "because it is missing"?

patricoferris:
From the port of vpnkit to the latest mirage packages https://github.com/moby/vpnkit/pull/646

hannesm:
Thanks, I'll merge. Please indicate if you need a release with this rather sooner than later.

Issue #534, opened 2026-02-17
Managing TCP window size and resource (memory) usage


edwintorok:
Memory usage of TCP applications in Mirage are a well-known problem [^1][^2]. There are workarounds (such as the restart-on-failure mode mentioned in the linked article), but they are not guaranteed to always work.

Defending against this with hardcoded limits is difficult, because packets can be of different size, and it is difficult to predict how much memory they'd use once decoded and handled by the Mirage stack (and user application).
Instead I proporse to (approximatively) track the memory used by a connection, and take actions based on that. The tracking doesn't have to be entirely precise, as long as the real memory used is within some constant factor of the desired target.

### mirage-tcpip could implement soft limits and window shrinking:

* allow the user to specify the desired memory target (especially important for unikernels, which may have hard limits on how much memory you can assign to them)

* have various parts of the code track (approximately) how much memory it is using (e.g. size of buffers, count of entries in various tables). This doesn't have to be entirely accurate (e.g. could be an overestimate), but at least it'd ensure that memory usage is bounded by at most a constant factor of the configured memory target.

* slow down the rate at which new connections are accepted if we get close to that memory limit, starting to drop packets once we actually hit it

* reduce the size of the TCP window for connections which have too high backlog of pending packets (this puts the burden on the other side to retry, see below about window shrinking and zero window). [^3]

* ensure keepalives are used in all corners cases. This is needed, because otherwise certain TCP states would require us to keep state forever , see [^6]

* Dropping packets should only be done as a last resort (hard limit), see [^4] for when dropping packets is not ideal (or even allowed by the RFC) 
 
### Hard limits

I have patches that implement some missing hard limits, and they improve the availability of mirage applications.
But for best results they should be combined with some soft limits as described above (I don't yet have patches for that).

Nevertheless hard limits are useful as a last line of defense against bugs or inaccuracies in the soft limit implementation.

### Testing

I have some code that triggers various issues in the Mirage stack, that can be used to test the effectiveness of various solutions.
For obvious reasons I won't be publishing that code.

### Similar approaches

I've successfully implemented a similar approach (tracking resource usage at runtime) to defend `oxenstored` against out of memory issues:
https://xenbits.xen.org/xsa/advisory-326.html

### Next steps

I propose to open some PRs which implement the fallback hard limits as a starting point (once we find a solution for https://github.com/mirage/mirage-tcpip/issues/533).

[...]

PR #535, opened 2026-02-17
(Configurable) hard limit on the number of open TCP connections

Firobe:
In line with https://github.com/mirage/mirage-tcpip/issues/534, this is Edwin's patch to put the hard limit to the number of TCP listens in progress, with some light modifications to make this configurable at `connect` time. It's intended as a last resort limit for this layer (but still before another limit in the lower layer: https://github.com/mirage/mirage-net-solo5/pull/49)

cc @edwintorok


Firobe:
Added a pair of commits to refresh the CI a bit :)

hannesm:
> Added a pair of commits to refresh the CI a bit :)

I guess that'd be best as a separate PR to be merged quickly.

So, you propose a "(Configurable) hard limit on the number of open TCP connections"? Your implementation seem to only care about the `listens` table, which are the partially opened server connections -- and not about the `channels` (fully established connections) neither the `connects` (partially opened client connections).

I wonder why, and wouldn't esp. for the `listens` table be an implementation of syncookies be sufficient (to avoid having to store much on the server side for any incoming SYN packet)? Is the mitigation primarily for syn floods, and in this case is it sufficient and the best solution?

edwintorok:
> I wonder why, and wouldn't esp. for the `listens` table be an implementation of syncookies be sufficient (to avoid having to store much on the server side for any incoming SYN packet)?

Although SYN cookies are a widely deployed solution, it can lead to silent data loss.
See [SYN cookies ate my dog, then told me I never had a dog, just a low MSS ](https://web.archive.org/web/20201026212800/https://kognitio.com/blog/syn-cookies-ate-my-dog-breaking-tcp-on-linux/)

Having the option to turn SYN cookies on or off might still be useful though, especially if the application has other ways of detecting missing or corrupted data (e.g. through TLS, so SYN cookies could be turned on by default when TLS is in use without the fear of undetectable data loss).

From the linked article:
> The problem of missing data can potentially affect any client-server TCP application in which SYN cookies are enabled on the server, the client speaks first after establishing a connection, the first packet it sends to the server is three bytes or smaller, and further data is sent before the server is expected to reply and before the client TCP stack resends the first packet.
> On Linux, if the client’s first packet is more than three bytes and a second packet is sent, this will not cause data to go silently missing from the TCP stream but it may cause the connection to be reset.

I was not actually aware of this until I started seeing some problems with SYN cookies myself, where legitimate traffic was getting stuck.

It can also lead to [performance issues](https://blog.cloudflare.com/syn-packet-handling-in-the-wild/#syn-cookies-and-tcp-timestamps), although if you only turn them on when you detect that you are being flooded that may be acceptable.

>  Is the mitigation primarily for syn floods, and in this case is it sufficient and the best solution?

SYN floods are one reason (and SYN cookies help there), but there is another attack which is not thwarted by SYN cookies. Just like how like SYN cookies avoid having to keep state server-side for connections, similar techniques can be used by an attacker client application, i.e. it is possible to write an almost entirely stateless TCP client capable of opening  a nearly unbounded amount of connections by fully completing a 3-way handshake without actually keeping any per-connection state (I've written such a stateless client myself, after finding a C implementation of this from 2008).

So it might be useful to limit both the amount of connections that are being established, and the amount of already established connections.

Firobe:
> > Added a pair of commits to refresh the CI a bit :)
> 
> I guess that'd be best as a separate PR to be merged quickly.

Done in https://github.com/mirage/mirage-tcpip/pull/536

> So it might be useful to limit both the amount of connections that are being established, and the amount of already established connections.

This makes me realize that the naming introduced in https://github.com/mirage/mirage/pull/1630#issuecomment-3920036437 is unclear. I refer to open connections, while we're actually talking about "pending" connections I guess. If we end up limiting both `listens` and `channels` I'll revise the naming.

PR #536, opened 2026-02-18
Fix CI

Firobe:
- raise example unikernel upper bound (no change needed)
- use OCaml 5.3, so bisect_ppx doesn't cause a solving conflict


hannesm:
thanks

Issue #537, opened 2026-02-24
127.0.0.0/8 is routed


reynir:
I was testing behavior in ocaml-dns using DNSvizor, and I noticed in tcpdump this:

    08:53:52.316379 IP 172.23.0.53.9603 > 127.0.2.1.53: 21607 [1au] A? null.reynir.dk. (43)

In other words a UDP packet (containing a DNS query) was sent to 127.0.2.1 over the interface. As I understand RFC 1122 addresses in 127.0.0.0/8 MUST NOT appear outside the host.

This might belong to arp.

hannesm:
I think this belongs here, esp. see src/ipv4/routing.ml - I guess also ipv6 should be checked for localhosts... At the same time, it may already support the (not yet approved, don't know where it is going) draft https://www.ietf.org/archive/id/draft-schoen-intarea-unicast-127-00.html ;)

PR #538, opened 2026-02-24
Error on attempts at writing martian packets

reynir:
If we try to write to a loopback address we error out. This is perhaps not ideal, but sending IP packets with loopback addresses is not allowed according to RFC 1122.

I'm not sure what to do with such packets. Ideally, they would instead be written to a loopback interface, but I don't think we have that.

Something similar should likely be done for ipv6, and there may be other addresses which we need to handle differently.

See also #537

### DIFF
--- a/src/ipv4/routing.ml
+++ b/src/ipv4/routing.ml
@@ -36,6 +36,8 @@ module Make(Log : Logs.LOG) (A : Arp.S) = struct
           Log.info (fun f -> f "IP.output: %a" A.pp_error e);
           Error `Local
       end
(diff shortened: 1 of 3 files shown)


hannesm (approved):
Thanks, looks good to me.

reynir:
The latest commit basically adds `if dst = ::1 then Error something else ` in the beginning of `write` in the IPv6 implementation. I'm not sure that's sufficient as there might be other places where routing is done. Changing the neighbor discovery code to implement this seemed less straightforward. So maybe we should drop that commit and think about other implementations?

Issue #498, opened 2022-10-24
API listen: be able to specify broadcast/multicast/unicast (or source IP address)


hannesm:
It would be great if e.g. a UDP.listen could specify "I'm only interested in multicast packets" or "I only want to receive unicast packets".

PR #540, opened 2026-08-04
Add multicast group membership (join/leave) to the IP and stack interfaces

### LINKED ISSUE #362: Multicast Listening on Unikernel?
Hi all,
I'm not sure this is an issue.
There is the possibility to listen on some multicast address when using mirage-tcpip inside an unikernel?
I see some ~group parameter in `dhcp_ipv4_stack` but I'm not sure is used for multicast groups.
I'm trying to modify the simple  TCP example to be able to listen also to UDP multicast.

config.ml:
```
open Mirage

let main = foreign ~packages: [package "duration"; package "fmt"] "Unikernel.Main"  (stackv4 @-> console @-> job)
let stack = dhcp_ipv4_stack ~group:"239.0.0.5" default_network
let () =
  register "services" [ main $ stack $ default_console]
```

unikernel.ml:
```
open Lwt.Infix

module Main (S: Mirage_types_lwt.STACKV4) (C: Mirage_types_lwt.CONSOLE) = struct
  let report_and_close flow pp e message =
    let ip, port = S.TCPV4.dst flow in
    Logs.warn
      (fun m -> m "closing connection from %a:%d due to error %a while %s"
          Ipaddr.V4.pp_hum ip port pp e message);
    S.TCPV4.close flow

  let task duration marker =
    ignore @@ OS.Time.sleep_ns (Duration.of_sec duration);
    Lwt.return @@ marker ^ marker

  let listener console udp port ~src ~dst ~src_port buf =
    let s = Format.asprintf "Data from %s to %s" (Ipaddr.V4.to_string src) (Ipaddr.V4.to_string dst) in
    ignore @@ C.log console s;
    let src' = (Ipaddr.V4 dst), port in
    let dst' = (Ipaddr.V4 src), src_port in
    S.UDPV4.write ~src_port:port ~dst:src ~dst_port:src_port udp buf >>= function
          | Ok () -> Lwt.return_unit
          | Error e -> Lwt.return_unit
          
 
  let rec echo flow =
    S.TCPV4.read flow >>= function
    | Error e -> report_and_close flow S.TCPV4.pp_error e "reading in Echo"
    | Ok `Eof -> report_and_close flow Fmt.string "end of file" "reading in Echo"
    | Ok (`Data buf) ->

      Lwt.choose[(task 1 "+");(task 2 "=")] >>= (
        fun x ->
          let b = Cstruct.of_string x
          in
          S.TCPV4.write flow b >>= function
          | Ok () -> echo flow
          | Error e -> report_and_close flow S.TCPV4.pp_write_error e "writing in Echo")

  let start s c =
   
    S.listen_udpv4 s ~port:7 (listener c (S.udpv4 s) 7);
    S.listen_tcpv4 s ~port:7 echo;
    S.listen s

end
```

These is a way to do that?
I need a new `direct stack` with UDP?

tmcgilchrist:
Adds join_multicast_group / leave_multicast_group to Tcpip.Ip.S and Tcpip.Stack.V4V6 so a stack can receive datagrams sent to a multicast group (the capability requested in #362 and #498) and what mDNS/DNS-SD (RFC 6762/6763) needs to answer queries on 224.0.0.251.
 
**Problem:** No way to join a group: the direct IPv4 stack drops any non-own-address/non-broadcast destination (static_ipv4.ml of_interest), the socket stack never issues IP_ADD_MEMBERSHIP. So a responder can announce but never receive.

**Change**
We add these functions to join and leave a multicast group:

``` ocaml
val join_multicast_group  : t -> ipaddr -> unit Lwt.t
val leave_multicast_group : t -> ipaddr -> unit Lwt.t
```

- Direct IPv4 (static_ipv4): tracks joined groups and accepts datagrams addressed to them at input. This is sufficient for **link-local** multicast (the `224.0.0.0/24` range, which includes mDNS's `224.0.0.251`) because switches flood that range and routers never forward it, so frames arrive without any on-wire membership announcement (ethernet already admits the multicast MAC). **Routable (non-link-local) groups** additionally require IGMP membership reports so the network forwards their traffic. Sending those, together with an IGMP implementation, which mirage-tcpip currently lacks, is left as future work.
- Dual stack (IPV4V6): routes to v4/v6; IPv6 MLD is a no-op for now.
- Socket UDP (udpv4v6_socket): IP_ADD_MEMBERSHIP on each bound fd and fds bound later, v4-on-dual-stack is best-effort (logged).
- Socket IP shim + v6 socket: no-op. Defaults preserve existing behaviour.

*Testing*: Wired into a MirageOS mDNS responder (dns-mdns-mirage) that I'm working on, it joins 224.0.0.251 at startup and is then discovered and resolved by stock Avahi, on both the unix target (direct stack over a tap) and Solo5/spt. Without this change the direct stack drops the queries and the responder is invisible.

Closes #362, #498.

### DIFF
--- a/src/core/ip.ml
+++ b/src/core/ip.ml
@@ -35,4 +35,6 @@ module type S = sig
   [@@ocaml.deprecated "this function will be removed soon, use [configured_ips] instead."]
   val configured_ips: t -> prefix list
   val mtu: t -> dst:ipaddr -> int
+  val join_multicast_group : t -> ipaddr -> unit Lwt.t
+  val leave_multicast_group : t -> ipaddr -> unit Lwt.t
 end
--- a/src/core/ip.ml
+++ b/src/core/ip.ml
@@ -37,4 +37,5 @@ module type S = sig
   val mtu: t -> dst:ipaddr -> int
   val join_multicast_group : t -> ipaddr -> unit Lwt.t
   val leave_multicast_group : t -> ipaddr -> unit Lwt.t
+  val multicast_groups : t -> ipaddr list
 end
--- a/src/core/ip.mli
+++ b/src/core/ip.mli
@@ -95,4 +95,15 @@ module type S = sig
   val mtu: t -> dst:ipaddr -> int
   (** [mtu ~dst ip] is the Maximum Transmission Unit of the [ip] i.e. the
       maximum size of the payload, not including the IP header. *)
+
+  val join_multicast_group : t -> ipaddr -> unit Lwt.t
+  (** [join_multicast_group t group] arranges for datagrams destined to the
+      multicast [group] to be accepted by {!input}.  For socket-based stacks
+      this performs an [IP_ADD_MEMBERSHIP], for the direct stack it records the
+      group so the input path stops discarding it.  Joining a non-multicast
+      address has no effect. *)
+
+  val leave_multicast_group : t -> ipaddr -> unit Lwt.t
+  (** [leave_multicast_group t group] reverses a previous
+      {!join_multicast_group}. *)
 end
--- a/src/core/ip.mli
+++ b/src/core/ip.mli
@@ -98,12 +98,23 @@ module type S = sig
 
   val join_multicast_group : t -> ipaddr -> unit Lwt.t
   (** [join_multicast_group t group] arranges for datagrams destined to the
-      multicast [group] to be accepted by {!input}.  For socket-based stacks
-      this performs an [IP_ADD_MEMBERSHIP], for the direct stack it records the
-      group so the input path stops discarding it.  Joining a non-multicast
-      address has no effect. *)
+      multicast [group] to be accepted by {!input}.  For the socket stacks
+      this performs an [IP_ADD_MEMBERSHIP].  For the direct stack it records
+      the group so the input path stops discarding it.  Joining an address
+      that is not multicast has no effect.
+
+      This delivers link local multicast, the [224.0.0.0/24] range (which
+      includes mDNS's [224.0.0.251]).  Switches flood that range and routers
+      never forward it, so the frames arrive without any membership
+      announcement on the wire.  Groups outside that range are not delivered,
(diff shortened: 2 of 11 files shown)


hannesm:
Thanks for your work. I'm no expert in multicast.

From the API side, I have some questions:
- join_multicast_group/leave_multicast_group - both result in `unit Lwt.t` -- why `Lwt.t`? and why `unit`? From the implementation it looks there can be failures (you provide a non-multicast address)
- there's no possibility to inspect the new field `mutable groups : Ipaddr.V4.Set.t` - may that be worth to add?
- if I understand this correctly, this PR provides functionality to join and leave multicast groups. Is something more needed to send multicast packets?
- is there some correlation (invariant/check) needed between multicast IP and mac address?

The comment you inserted "IGMP membership reports are only needed for non link-local groups and are left as future work" -- is it worth to document this in the API?

How does this solve #498? How can a listen call now specify to only accept multicast packets? (I don't remember what use case I had in mind when I opened #498 to be honest)

tmcgilchrist:
> Why `unit Lwt.t`?

Yes the `unit` hides errors, not that there are many right now. The only real error is the socket IP_ADD_MEMBERSHIP failure is logged when it could also be returned as an `error`. In future a real IGMP implementation would send membership reports with the join/leave calls which would need to be Lwt.t. What about making the return `(unit, error) result Lwt.t`? Do you prefer extending `Ip.error` with a multicast variant, or a local error type?

For now I'm leaving the IGMP implementation as future work, I don't strictly need it for mDNS. 

> No way to inspect groups.

Yes, something like `val multicast_groups : t -> ipaddr list.` seems right.

> Is something more needed to send multicast packets?

No, sending multicast packets already works. This PR is receive-only.

> How does this solve #498? How can a listen specify multicast-only?

This PR adds group membership, which is how I interpreted #498  when I read "I'm only interested in multicast packets". I assumed you wanted more control over which packets you needed to process, which is partially there with the filtering being done in this change but only on the receive side for multicast.

PR #541, opened 2026-08-06
raise test unikernel's mirage upper bound

tmcgilchrist:
mirage 4.11 is released, so `mirage configure` in examples/unikernel aborts with "Version mismatch: required is mirage < 4.11.0, but 4.11.2 is installed", failing the unikernel-example (unix/virtio/qubes) jobs on main since the 4.11.0 release.


hannesm:
thanks

